Issue #29, opened 2017-09-16
Enhance object cache for policy.


k82cn:
For policy, we need a snapshot of cluster, e.g. Pod, Node, to calculate how many resource should be allocated to a tenant. This task is used to provide such kind of cache and related helper functions.

Sub-Tasks:

- [x] `Resource` type and related helper functions, e.g. `Resource.Add`
- [x] Pod and Node cache
- [x] Consumer cache

k82cn:
For Pod/Node cache, we only need to cache resource related info: cpu and memory, instead of whole Pod object.

Refer to https://github.com/k82cn/kube-resmgr/blob/master/pkg/util/resource.go for `Resource` type.

k82cn:
Close this as all sub-tasks are finished.

PR #38, opened 2017-09-18
Addressing issue #29 Helper functions for resource cache

swiftdiaries:
Re-submission of PR, added resource_info.go with helper functions.


k82cn:
@swiftdiaries , can you build this PR against master? I'm going to delete 'release-0.1-working' branch.

swiftdiaries:
Sure thing. Will do.

pkg/schedulercache/resource_info.go
```diff
@@ -16,5 +16,71 @@ limitations under the License.
...
+
+func (r *Resource) Add(rr *Resource) *Resource {
+	r.MilliCPU += rr.MilliCPU
+	r.Memory += rr.Memory
+	return r
+}
+
+func (r *Resource) Sub(rr *Resource) *Resource {
+	r.MilliCPU -= rr.MilliCPU
+	r.Memory -= rr.Memory
```
jinzhejz: suggest check the case `r.MilliCPU < rr.MilliCPU` and `r.Memory < rr.Memory`

pkg/schedulercache/resource_info.go
```diff
@@ -16,5 +16,71 @@ limitations under the License.
...
+	return r
+}
+
+func (r *Resource) Less(rr *Resource) bool {
+	return r.MilliCPU < rr.MilliCPU && r.Memory < rr.Memory
```
jinzhejz: this may return `false` for `A.Less(B)` and `B.Less(A)`, such as `A(MilliCPU:2, Memory:1)`, `B(MilliCPU:2, Memory:1)`

pkg/schedulercache/resource_info.go
```diff
@@ -16,5 +16,71 @@ limitations under the License.
...
+}
+
+func (r *Resource) LessEqual(rr *Resource) bool {
+	return (r.MilliCPU < rr.MilliCPU || math.Abs(rr.MilliCPU-r.MilliCPU) < 0.01) &&
+		(r.Memory < rr.Memory || math.Abs(rr.Memory-r.Memory) < 1)
```
jinzhejz, line 97: suggest remove `Less()` and `LessEqual()` temporarily, or compare `MilliCPU` and `Memory` independently

k82cn:
xref #29

pkg/schedulercache/resource_info.go
k82cn: I think that's what we want. When allocating resource to Pod, we need to know whether we can launch Pod with resources.

jinzhejz: get the point :)

pkg/schedulercache/resource_info.go
swiftdiaries: In the case when both (r.MilliCPU and r.Memory) are less than their counterparts, what should I do with the result? Would it be ok to print out that the resource is not available?

pkg/schedulercache/resource_info.go
swiftdiaries: Meaning wherever we do need to compare the two Resources, we do it manually each time?

pkg/schedulercache/resource_info.go
jinzhejz: sure, I think the print log is enough now

pkg/schedulercache/resource_info.go
jinzhejz: as @k82cn mentioned in above comments, I misunderstand the usage of `Less()` and `LessEqual()`, we can keep the current behavior

pkg/schedulercache/resource_info.go
k82cn: In MVP, I'd suggest to panic it directly; and then, we need to know how to handle this exception.

For the panic, please create a wrapper func; so we can only log error info in future without go through the whole code.

swiftdiaries: Got it :) Thanks for the suggestions!

pkg/schedulercache/resource_info.go
swiftdiaries: Ok :) 👍

k82cn:
@swiftdiaries , are you still working on this PR?

swiftdiaries:
I have edited according to suggested comments. Is there anything else I
need to add?

pkg/schedulercache/resource_info.go
```diff
@@ -16,5 +16,81 @@ limitations under the License.
...
+		MilliCPU: r.MilliCPU,
+		Memory:   r.Memory,
+	}
+	return clone
+}
+
+var minMilliCPU float64 = 10
+var minMemory float64 = 10 * 1024 * 1024
+
+//Changed NewResource to account for single resource requests
```
k82cn: What's purpose of this comments?

pkg/schedulercache/resource_info.go
```diff
@@ -16,5 +16,81 @@ limitations under the License.
...
+	return r.MilliCPU < minMilliCPU && r.Memory < minMemory
+}
+
+func (r *Resource) Add(rr *Resource) *Resource {
+	r.MilliCPU += rr.MilliCPU
+	r.Memory += rr.Memory
+	return r
+}
+
+//Added panic to show when Resource is insufficient
```
k82cn: a TODO?

pkg/schedulercache/resource_info.go
```diff
@@ -16,5 +16,81 @@ limitations under the License.
...
+
+func (r *Resource) Sub(rr *Resource) *Resource {
```
k82cn: add comments for the behaviours of Sub.

pkg/schedulercache/resource_info.go
```diff
@@ -16,5 +16,81 @@ limitations under the License.
...
+	if r.Less(rr) == false {
+		r.MilliCPU -= rr.MilliCPU
+		r.Memory -= rr.Memory
+		return r
+	}
+	panic("Resource is not sufficient to do operation: Sub()")
```
k82cn, line 88: Add a wrapper func for it; we'll remove panic when it's stable.

swiftdiaries:
Oh those were addressing review comments.

swiftdiaries:
Sure thing. I'll add them and update. Thanks for the review.

k82cn:
And please
1. remove the dependency on `vendor/github.com/ugorji/*`
2. merge those 4 commits into one  (here's the command I use: `git reset`, `git commit`, `git push --force`)

swiftdiaries:
Will do that. Thanks for the pointers.

k82cn:
1. `Godeps/Godeps.json` is conflict
1. why introduce `github.com/ugorji/`?
1. Please follow golang's comment style :).

k82cn:
@swiftdiaries , I'd like to get this PR merged after removing the diff about vendor; would you help on that.

swiftdiaries:
Yeah, on it. Will get it done.

swiftdiaries:
Closing and creating a new one.

PR #41, opened 2017-09-20
Add integration test framework and update vendor for it

jinzhejz:
Add integration test framework and update vendor for it. will create another PR for integration test case after this PR is merged


hack/lib/golang.sh
```diff
@@ -148,14 +150,9 @@ readonly KUBE_TEST_BINARIES=("${KUBE_TEST_TARGETS[@]##*/}")
 readonly KUBE_TEST_BINARIES_WIN=("${KUBE_TEST_BINARIES[@]/%/.exe}")
 # If you update this list, please also update build/BUILD.
 readonly KUBE_TEST_PORTABLE=(
-  test/e2e/testing-manifests
   test/kubemark
   federation/develop
```
k82cn, line 154: federation should be not necessary.

jinzhejz:
close it to address comments and refine commit format

k82cn:
@jinzhejz , please try to update this PR instead of a new one; we also need to trace reviewer's comments.

PR #42, opened 2017-09-20
Init integration test framework and update vendor

jinzhejz:
Init integration test framework and update vendor


k82cn (approved):
let's get this PR merged firstly. I did not go through the detail, suppose it was just copied from upstream :).

jinzhejz:
yes, most of the code is from upstream :)

PR #48, opened 2017-10-09
refine quotamanage and proportion as design doc

jinzhejz:
Two changes

1. proportion scheduler will update scheduling result into `Queue`
2. quota manager fetch scheduling result from `Queue` and update it into `Resource Quota`


jinzhejz:
Close this PR first and create a new one to pass CI

PR #51, opened 2017-10-10
Added policy factory and parameter.

k82cn:
Added policy factory and parameter, so user can choose different policy when start the arbitrator.


jinzhejz (approved):
lgtm

Issue #30, opened 2017-09-16
Queue API & Controller


k82cn:
`Queue` is an important API object in kube-arbitrator. This issue is used to trace all related features.

- [ ] Design doc of `Queue`

PR #28, opened 2017-09-15
WIP: Queue API design doc.

k82cn:
xref https://github.com/kubernetes-incubator/kube-arbitrator/issues/30

PR #57, opened 2017-10-16
Queue API design doc

jinzhejz:
xref #30 

Can't change PR #28 , so I create this one to update API doc.


doc/design/queue_api.md
```diff
@@ -0,0 +1,382 @@
...
+	Dump() *CacheSnapshot
+}
+
+type CacheSnapshot struct {
+	Pods       []*PodInfo
+	Nodes      []*NodeInfo
+	Queues     []*QueueInfo
+}
+```
+
```
hanghliu: If schedulercache could be shared by queuejob controller, or potential other additional controllers, various "snapshot" may be needed. For example,  a "QueueJobCacheSnapshot" would contains PodInfo/QueueInfo/QueuejobInfo.

doc/design/queue_api.md
```diff
@@ -0,0 +1,382 @@
...
+|                                                                        |
+|    Allocated:        Used:                Pods:                        |
+|      cpu: 6            cpu: 5               pod-1: cpu=2 memory=1Gi    |
+|      memory: 18Gi      memory: 3Gi          pod-2: cpu=2 memory=1Gi    |
+|                                             pod-3: cpu=1 memory=1Gi    |
+--------------------------------------------------------------------------
+
+There are 9 CPUs totally. 3 CPUs are allocated to Queue-1 and 6 CPUs are allocated to Queue-2.
+However, Queue-1 is overused due to some race condition. So pod-1 will be chosen to terminate. The pod is randomly selected currently.
+--------------------------------------------------------------------------
```
hanghliu: Is it possible to provide a callback function to choose Pods to be killed?

jinzhejz: yes, we can enhance the pod selection, such as provide some default selection functions or a callback.
Currently, the pod is chosen randomly.

doc/design/queue_api.md
jinzhejz: I think we may add `QueuejobInfo` into `CacheSnapshot`. Controller dumps all snapshots and uses related information.
Maybe we can separate the snapshot later if dumping extra information is too heavy.

doc/design/queue_api.md
```diff
@@ -0,0 +1,382 @@
+# Queue API
+
+@k82cn, 9/16/2017
```
k82cn: also added yourself as author.

doc/design/queue_api.md
```diff
@@ -0,0 +1,382 @@
...
+* Update `Allocated ` information in `QueueStatus` into Resource Quota
+
+### Scheduler Cache
+
+Scheduler Cache periodically fetches all Node/Pod/Queue information in the cluster from API server. That inforamtion will only be stored in memory and not persisted on disk.
+
+It provides two interfaces `Run()` and `Dump()`
+
+* `Run()` to trigger cache to periodically fetch Node/Pod/Queue information from API server
+* `Dump()` create `CacheSnapshot` for policy
```
k82cn: I'd like to call it `Snapshot` :).

jinzhejz:
Remove details of preemption from the API doc, and will provide MVP doc of preemption later.

doc/design/queue_api.md
```diff
@@ -0,0 +1,178 @@
...
+	Preprocessing(queues map[string]*schedulercache.QueueInfo, pods []*schedulercache.PodInfo) (map[string]*schedulercache.QueueInfo, error)
+
+	// PreemptResources preempt resources between queue
+	// Deserved/Allocated/Used/Preempting will be adjusted
+	PreemptResources(queues map[string]*schedulercache.QueueInfo) error
+}
+```
+
+### Queuejob
+The proportion policy just used Queue `weight` to allocate resources now,  it does not consider the queue job information, such as priority, resources requirement or some other factors. And we need provide some complex scheduling strategy to consider these queue job information.
```
hanghliu, line 185: Would the scheduling strategy for queuejob be addressed in an additional separate design doc?

doc/design/queue_api.md
```diff
@@ -0,0 +1,178 @@
...
+    Used        ResourceList
+    // The resources that are preempting for Queue
+    Preempting  ResourceList
+}
+```
+
+## Function Detail
+
+### Workflow
+![workflow](../images/workflow.jpg)
```
hanghliu, line 72: In workflow.jpg diagram, 
1) I think the queuejob controller might be also added, with same position as the "Quota Manager". It will watch the queue changes as well, but its action should be adjust queuejob size(scale up/down) in the queue changed. 
2) The queuejob objects should also be added into the apiserver, which is the action targets of queuejob controller.

doc/design/queue_api.md
jinzhejz: yes, this is just for API design, we will provide a separate design doc for more scheduling strategy, and I think some discussion may be needed before it.

doc/design/queue_api.md
jinzhejz: got it, could you provide a draft diagram so I can add it into workflow.jpg, or you may update it after this PR is merged.

doc/design/queue_api.md
hanghliu: got it.

doc/design/queue_api.md
hanghliu: I can send you a draft diagram through email.

jinzhejz: Update the diagram.

I have a little confuse about the main function of Queue Controller, according to my understanding, it will maintain(add/update/delete) the queue job information(such as priority, resources requirement) in API server. And queue job resources will be calculated/allocated by scheduling policy, right?

k82cn:
@kubernetes-incubator/design-reviewer-kube-arbitrator 

demo video of this design: https://youtu.be/VkT6_Zqesz8

PR #84, opened 2017-11-06
use runtime.RawExtension to represent queuejob sub-resource

hanghliu:
Use runtime.RawExtension to represent the sub-resource in queuejob API  object definition, thus it can support more resource types than pod.


jinzhejz (commented):
suggest removing the commented out codes

jinzhejz (approved):
lgtm

PR #86, opened 2017-11-08
Add queuejob reference manager

hanghliu:
Add queuejob reference manager to manage the membership between queuejob resources and its members.

### DIFF
--- a/pkg/controller/queuejob_controller.go
+++ b/pkg/controller/queuejob_controller.go
@@ -286,9 +286,12 @@ func (qjm *QueueJobController) syncQueueJob(key string) error {
 
 	if job.Spec.AggrResources.Items != nil {
 		for i := range job.Spec.AggrResources.Items {
-			qjm.refManager.AddTag(&job.Spec.AggrResources.Items[i], func() string {
+			err := qjm.refManager.AddTag(&job.Spec.AggrResources.Items[i], func() string {
 				return strconv.Itoa(i)
 			})
+			if err != nil {
+				return err
+			}
 
 		}
 		var result qjobv1.QueueJob
--- a/pkg/controller/queuejob_controller.go
+++ b/pkg/controller/queuejob_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


pkg/controller/queuejobresources/pod/pod.go
```diff
@@ -176,11 +179,12 @@ func (qjrPod *QueueJobResPod) scaleUpQueueJobResource(resIndex int, diff int32,
 		return err
 	}
 
-	if template.Template.Labels == nil {
-
-		template.Template.Labels = map[string]string{}
+	//TODO: need set reference after Pod has been really dded
```
jinzhejz: seems a typo `dded`

pkg/controller/queuejob_controller.go
```diff
@@ -279,8 +285,19 @@ func (qjm *QueueJobController) syncQueueJob(key string) error {
 	}
 
 	if job.Spec.AggrResources.Items != nil {
-		for i, ar := range job.Spec.AggrResources.Items {
-			qjm.qjobResControls[ar.Type].Sync(i, sharedJob, &ar)
+		for i := range job.Spec.AggrResources.Items {
+			qjm.refManager.AddTag(&job.Spec.AggrResources.Items[i], func() string {
```
jinzhejz: suggest adding error handling for `AddTag` when the return value is not nil

pkg/controller/queuejobresources/pod/pod.go
```diff
@@ -176,11 +179,12 @@ func (qjrPod *QueueJobResPod) scaleUpQueueJobResource(resIndex int, diff int32,
...
+	tmpPod := v1.Pod{}
+	qjrPod.refManager.AddReference(qjobRes, &tmpPod)
```
jinzhejz: suggest adding error handling for `AddReference` when the return value is not nil

pkg/controller/queuejob_controller.go
hanghliu: ok

pkg/controller/queuejobresources/pod/pod.go
hanghliu: ok

pkg/controller/queuejobresources/pod/pod.go
hanghliu: Yes, it should be added

jinzhejz (approved):
lgtm

PR #87, opened 2017-11-13
assign left resources which can not be allocated by weight

jinzhejz:


pkg/policy/proportion/proportion.go
```diff
@@ -103,6 +110,29 @@ func (ps *proportionScheduler) Allocate(
...
+		}
+		if leftMEM > 0 {
+			leftMEM -= 1
+			result := resList.Resources["memory"].DeepCopy()
+			result.Add(*resource.NewQuantity(1, resource.BinarySI))
+			queue.Queue().Status.Deserved.Resources["memory"] = result
+		}
+	}
+	if leftCPU > 0 || leftMEM > 0 {
+		glog.Errorf("fail to assign the left resources, left cpu %d, left memory %d", leftCPU, leftMEM)
```
k82cn: we should make sure all resources are allocated instead of an error message.

jinzhejz: The `leftCPU` and `leftMEM` is less than the size of `allocatedResult`, so all of them can be allocated in above loop. I will remove the error message.

k82cn (approved):

PR #89, opened 2017-11-14
support resource request in queue

jinzhejz:
Add a new policy `maxmin` to support allocate resources to each queue by weight and resource request(CPU and memory) by using `max-min weighted fairness` algorithm


k82cn:
Can we only enhance proportion policy to handle resource request?

jinzhejz:
yes, we can.

The current difference between proportion and maxmin:
1. In proportion, the queue can be allocated resources when weight is configured.
2. In maxmin, the queue can be allocated resource when both weight and resource request is configured. Only configure weight of a queue will make the queue get no resoures.

jinzhejz:
Enhance proportion policy for resource request in queue

k82cn (approved):

PR #95, opened 2017-12-04
Only schedule pending/running pods

jinzhejz:


k82cn (dismissed):

pkg/controller/queue_controller.go
```diff
@@ -57,10 +57,12 @@ func (q *QueueController) Run(stopCh <-chan struct{}) {
 func (q *QueueController) runOnce() {
 	glog.V(4).Infof("Start scheduling ...")
 	defer glog.V(4).Infof("End scheduling ...")
+
 	snapshot := q.cache.Dump()
+	allPods := q.allocator.Initialize(snapshot.Pods)
```
k82cn: `Initialize` is used for policy plugin initialization, e.g. load `.so`. Prefer to add this into `Group (...)`.

jinzhejz: got it

Issue #69, opened 2017-10-28
Enhancement: multiple strategy of pod selection


jinzhejz:
The policy will terminate pod to release resources if the `Queue` is underused (Used > Allocated). Now the pod is selected randomly. Need support more strategy of pod selection according to some factors, such as

- Running time
- Priority
- Labels

elaron:
In which circumstances a queue which will be in the state of **underused**? How can the _used_ resource greater than _allocated_ resource?

jinzhejz:
kube-arbitrator allocates resources to queues by weight, if there are no enough tasks in a queue, it will cause a queue underused.

Currently, kube-arbitrator will update scheduling result to `ResourceQuota` periodically, which means the allocated resources.
There is a time window between two updates and the user can also create pods in the time window. If the second update is less than the first update, it may cause the used resource greater than the allocated resource.

PR #99, opened 2017-12-12
preempt pods from low priority to high priority

jinzhejz:
This PR to fix a case for #69 

Kube-arbitrator will get label "preemptionrank" of pod if it exists. When preemption happens, kube-arbitrator chooses pod with lower "preemptionrank" to terminate first.

The label "preemptionrank" will be treated as 0 by kube-arbitrator if it doesn't exist.

### DIFF
--- a/pkg/policy/preemption/podslice.go
+++ b/pkg/policy/preemption/podslice.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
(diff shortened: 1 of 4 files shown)


jinzhejz:
Kube-arbitrator will sort pods as following conditions:
1. preemption rank. The lower rank pod will be terminated first.
2. pod status. The pending pod will be terminated first.
3. pod running time. The pod with shorter runtime will be terminated first.

k82cn (approved):

PR #120, opened 2018-01-05
Fix header format.

ConnorDoyle:


doc/design/queue_api.md
```diff
@@ -182,4 +182,4 @@ type Interface interface {
 ```
 
 ### Queuejob
-The proportion policy just used Queue `weight` to allocate resources now,  it does not consider the queue job information, such as priority, resources requirement or some other factors. And we need provide some complex scheduling strategy to consider these queue job information.
\ No newline at end of file
+The proportion policy just used Queue `weight` to allocate resources now,  it does not consider the queue job information, such as priority, resources requirement or some other factors. And we need provide some complex scheduling strategy to consider these queue job information.
```
ConnorDoyle, line 185: Think this was just a newline change; side effect of github's online editor.

PR #124, opened 2018-01-08
update pod info in consumer

jinzhejz:
* add/update consumer information in cache when add/delete pod
* update ut case for cache and drf
* exposed `SchedulerCache` for ut

### DIFF
--- a/pkg/policy/drf/drf_test.go
+++ b/pkg/policy/drf/drf_test.go
@@ -22,8 +22,11 @@ import (
 	"os"
 	"testing"
 
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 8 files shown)


pkg/policy/drf/drf_test.go
```diff
@@ -37,159 +37,5 @@ func init() {
...
-					pk := fmt.Sprintf("%v/%v", p.Namespace, p.Name)
-					if p.NodeName != test.expected[pk] {
-						t.Errorf("case %d (%s): %v/%v expected %s got %s",
-							i, test.name, p.Namespace, p.Name, test.expected[pk], p.NodeName)
-					}
-				}
-			}
-		}
-	}
+	// TODO(jinzhejz): build consumer information in cache
```
k82cn: please do not delete UT directly.

k82cn (approved):

PR #130, opened 2018-01-16
cancel all alloc decision processing before policy run

jinzhejz:


pkg/controller/policy_controller.go
```diff
@@ -93,7 +93,7 @@ func (pc *PolicyController) runOnce() {
 	glog.V(4).Infof("Start scheduling ...")
 	defer glog.V(4).Infof("End scheduling ...")
 
-	// TODO(k82cn): cancel all alloc decision processing firstly.
+	pc.resetPodSets()
```
k82cn: rename to `cancelAllocDecisionProcessing`

k82cn (approved):

PR #131, opened 2018-01-16
use evict pod instead of delete pod

jinzhejz:


pkg/controller/policy_controller.go
```diff
@@ -134,8 +135,14 @@ func (pc *PolicyController) processAllocDecision() {
...
-				// TODO(k82cn): it's better to use /eviction instead of delete to avoid race-condition.
-				if err := pc.kubeclient.CoreV1().Pods(p.Namespace).Delete(p.Name, &metav1.DeleteOptions{}); err != nil {
+				// use /eviction instead of delete to avoid race-condition.
+				evict := &policyv1.Eviction{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      p.Name,
+						Namespace: p.Namespace,
+					},
+				}
+				if err := pc.kubeclient.CoreV1().Pods(p.Namespace).Evict(evict); err != nil {
```
k82cn, line 153: Does `Evict` work as we expected? If not, I'd like to keep this TODO and raise a requirement to upstream.

jinzhejz: `Evict` will terminate the pod directly. It will not change the pod to pending status for next scheduling.

k82cn: If Pod was deleted (similar behaviour to Delete), I'd like to keep TODO here.

jinzhejz:
Close this PR due to `Evict` does not work as expected. Will raise a requirement later.

PR #133, opened 2018-01-22
Add PodDisruptionBuget into cache

jinzhejz:
only support integer `MinAvailable` in `PodDisruptionBuget`.

percentage `MinAvailable`, integer/percentage `MaxUnavailable` will be supported later.

### DIFF
--- a/pkg/cache/consumer_info.go
+++ b/pkg/cache/consumer_info.go
@@ -17,9 +17,13 @@ limitations under the License.
 package cache
 
 import (
-	arbv1 "github.com/kubernetes-incubator/kube-arbitrator/pkg/apis/v1"
-
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/golang/glog"
+	arbv1 "github.com/kubernetes-incubator/kube-arbitrator/pkg/apis/v1"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/cache/consumer_info.go
```diff
@@ -93,6 +97,42 @@ func (ci *ConsumerInfo) RemovePod(pi *PodInfo) {
...
+			continue
+		}
+		selector, err := metav1.LabelSelectorAsSelector(pi.Pdb.Spec.Selector)
+		if err != nil {
+			glog.V(4).Infof("LabelSelectorAsSelector fail for pdb %s", pi.Name)
+			continue
+		}
+		if selector.Matches(labels.Set(ps.Labels)) {
+			ps.PdbName = ""
+			ps.MinAvailable = 1
```
k82cn: should be 0

pkg/cache/consumer_info.go
```diff
@@ -93,6 +97,42 @@ func (ci *ConsumerInfo) RemovePod(pi *PodInfo) {
...
+	for _, ps := range ci.PodSets {
+		if len(ps.PdbName) != 0 {
+			continue
+		}
+		selector, err := metav1.LabelSelectorAsSelector(pi.Pdb.Spec.Selector)
+		if err != nil {
+			glog.V(4).Infof("LabelSelectorAsSelector fail for pdb %s", pi.Name)
+			continue
+		}
+		if selector.Matches(labels.Set(ps.Labels)) {
```
k82cn, line 112: Add some TODO here, we will not handle the PDB that cross different PodSet.

k82cn (approved):

PR #136, opened 2018-01-25
fix race condition between cache and policy

jinzhejz:
There is a race condition between policy and cache, pod status in cache may be not consistent with last policy schedule result.

`podSetInfo` will go through Pending pods to calculate the assigned Pod number to reduce the frequency. However, the issue will not be fixed totally by this PR.

The following ways may fix this issue, need some investigation
* Policy will not start next schedule till it updates all scheduler result to api-server
* When policy assign resources, consider the last schedule result, policy needs to record the last scheduler result

### DIFF
--- a/pkg/controller/policy_controller.go
+++ b/pkg/controller/policy_controller.go
@@ -85,7 +85,7 @@ func (pc *PolicyController) Run(stopCh <-chan struct{}) {
 	go pc.cache.Run(stopCh)
 	pc.cache.WaitForCacheSync(stopCh)
 
-	go wait.Until(pc.runOnce, 2*time.Second, stopCh)
+	go wait.Until(pc.runOnce, 4*time.Second, stopCh)
 	go wait.Until(pc.processAllocDecision, 0, stopCh)
 }
 
--- a/pkg/policy/drf/drf.go
+++ b/pkg/policy/drf/drf.go
@@ -50,9 +50,10 @@ func (drf *drfScheduler) Allocate(consumers []*cache.ConsumerInfo, nodes []*cach
 	pq := util.NewPriorityQueue()
 
 	total := cache.EmptyResource()
-
+	nodesMap := make(map[string]*cache.NodeInfo)
 	for _, n := range nodes {
 		total.Add(n.Allocatable)
+		nodesMap[n.Name] = n
 	}
 
 	for _, c := range consumers {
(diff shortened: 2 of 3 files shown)


k82cn (changes requested):
please add more detail about the root cause :)

jinzhejz:
#### The brief workflow of each thread/component:

##### T1. Cache thread:
  
  1. Update Pod/Node/Consumer information from api-server

##### T2.Policy - resource alloc thread:
  
  1. Get Pod/Node/Consumer snapshot from Cache
  2. Allocate resources for Pod/PodSet in the cache snapshot
    2.1. Assign MinAvailable to each PodSet one by one
    2.2. Assign left resources to each PodSet by DRF
  3. Add allocate resources result in a queue, name it as `AllocationQueue`

##### T3.Policy - process alloc decision thread:
  
  1. Get the scheduler result from `AllocationQueue` and update it into api-server

##### T4.Kubelet
  
  1. Get Pod information from api-server, start/stop it and then update status to api-server


### Case 01 workflow (this PR try to fix)
1. In step T2.2.1: Policy allocate MinAvailable to PodSet
2. In step T2.3: Policy add allocate result to `AllocationQueue`  (Pods still pending)
3. In step T3.1: Policy update allocate result to api-server (Pods still pending, with Nodename)
4. In step T1.1: Cache update information from api-server in step T1.1 (Pods still pending, with Nodename)
5. In step T2.1: Policy get cache snapshot (Pods still pending, with Nodename)
6. In step T2.2.1: Policy allocate MinAvailable to PodSet, however, the policy doesn't calculate the number of pending pods with Nodename, still get pods from pending queue one by one.
7. In step T3.1: Policy update allocate result to api-server (more Pods still pending, with Nodename)
8. In step T4.1: Kubelet start pod 

### Case 02 workflow
1. In step T2.2.1: Policy allocate MinAvailable to PodSet 
2. In step T2.3: Policy add allocate result to `AllocationQueue` (Pods still pending)
3. In step T3.1: Policy update allocate result to api-server (Pods still pending, with Nodename)
4. In step T2.1: Policy get cache snapshot (Pods still pending)
5. In step T1.1: Cache update information from api-server (Pods still pending, with Nodename)
6. In step T2.2.1: Policy allocate MinAvailable to PodSet, however, all the pods in the pending queue without nodename, it may cause policy allocate more MinAvailable in this step.
7. In step T3.1: Policy update allocate result to api-server (more Pods still pending, with Nodename)
8. In step T4.1: Kubelet start pod

k82cn:
My understanding is that: there's race condition between `cancelAllocDecisionProcessing` and `Snapshot`. For this case, is it possible to assume the pod in cache before binding? please refer to kube-scheduler for it.

jinzhejz:
I think not only the race condition between `cancelAllocDecisionProcessing` and `Snapshot`, but also between `SchedulerCache.UpdatePod` and `SchedulerCache.Snapshot`.

case01/02 show the race condition between `SchedulerCache.UpdatePod` and `SchedulerCache.Snapshot`. All alloc results are updated to api-server in _step 3_ in case01/02

I will refer to kube-arbitrator to get more details.

k82cn:
btw, for such kind of issue, please open an issue for discussion instead of a PR. For tiny fix, it's OK to open RP directly; but for such complex issue, let's trace the discussion in the issue :).

jinzhejz:
Close this PR first, and open an issue #137 for discussion,

PR #151, opened 2018-01-30
build priority queue after assign minAvailable

jinzhejz:
Build priority queue for DRF after assign minAvailable, otherwise, the podset priority is incorrect.


k82cn:
@jinzhejz , would you help to open an issue to add UT for this case?

PR #163, opened 2018-02-07
gang scheduling for PodSet

jinzhejz:
The brief scheduling process for `minAvailable`:

1. Policy assign `minAvailable` for each PodSet first
2. If all PodSet get their `minAvailable` in step 1, then policy assign left resources to these PodSet by DRF. 
3. If some of PodSets could not get their `minAvailable` in step 1, the policy will not assign any resources to PodSet. The unassigned resources will be kept by policy.

### DIFF
--- a/pkg/batchd/policy/drf/drf.go
+++ b/pkg/batchd/policy/drf/drf.go
@@ -58,10 +58,12 @@ func (drf *drfScheduler) Allocate(queues []*cache.QueueInfo, nodes []*cache.Node
 		}
 	}
 
-	sort.Sort(dq)
 	// assign MinAvailable of each podSet first by chronologically
+	sort.Sort(dq)
+	meetAllMinAvailable := true
 	for _, q := range dq {
 		psi := q.Value.(*podSetInfo)
+		assignedPods := make(map[string]string)
 		for {
 			if psi.meetMinAvailable() {
 				glog.V(3).Infof("MinAvailable of podset %s/%s is met, assign next podset",
@@ -84,6 +86,8 @@ func (drf *drfScheduler) Allocate(queues []*cache.QueueInfo, nodes []*cache.Node
 
 					assigned = true
 
+					assignedPods[p.Name] = node.Name
+
 					glog.V(3).Infof("assign <%v/%v> to <%s>: available <%v>, request <%v>",
 						p.Namespace, p.Name, p.NodeName, node.Idle, p.Request)
 					break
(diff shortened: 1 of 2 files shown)


pkg/batchd/policy/drf/drf.go
```diff
@@ -98,6 +99,27 @@ func (drf *drfScheduler) Allocate(queues []*cache.QueueInfo, nodes []*cache.Node
 				break
 			}
 		}
+
+		// If policy could not assign minAvailable for a PodSet, then recycle assigned resources
+		if !psi.meetMinAvailable() {
+			assignedPending := psi.resetAndGetAssignedPod()
```
k82cn: I think we need to trace what we had assign in this loop instead of reset all assigned pods.

jinzhejz: got it

k82cn:
please add UT for this :)

pkg/batchd/policy/drf/drf.go
```diff
@@ -98,6 +102,32 @@ func (drf *drfScheduler) Allocate(queues []*cache.QueueInfo, nodes []*cache.Node
...
+				}
+			}
+			meetAllMinAvailable = false
+		}
+	}
+
+	// If minAvailable of all PodSets are meet, then assign left resources to these PodSets by DRF
+	// Otherwise, policy only allocate minAvailable for each PodSet. The left resources will not be
+	// assigned to the left PodSet.
+	if !meetAllMinAvailable {
```
k82cn, line 126: That's not necessary to me.

pkg/batchd/policy/drf/drf.go
```diff
@@ -98,6 +102,32 @@ func (drf *drfScheduler) Allocate(queues []*cache.QueueInfo, nodes []*cache.Node
 				break
 			}
 		}
+
+		// If policy could not assign minAvailable for a PodSet, then recycle assigned resources in
+		// this loop even if the PodSet was assigned resources before it is to handle a case that PodSet
+		// meet minAvailable in the past and then become under minAvailable due to some pods finish
+		if !psi.meetMinAvailable() {
+			for podName, nodeName := range assignedPods {
+				pi := psi.resetAndGetAssignedPod(podName)
```
k82cn, line 108: prefer to use "flag or mark", "commit if meet, otherwise discard" steps; instead of "add", "sub" :).

k82cn: And we may add some new func/method to PodSetInfo.

jinzhejz:
Close this PR and will separate it into several smaller PRs

PR #168, opened 2018-02-13
*: Add golinter

### LINKED ISSUE #64: Add golint check for source code
Now there is gofmt check by `hack/update-gofmt.sh`. 

Need to add golint check for source code and refine code according to golint result. Such as, adding a new script as `hack/update-golint.sh` for the check.

- [X] Add golint tools into this repo (#168)
- [ ] Enable golint check, at least policy directory

gaocegege:
Close #64 

Signed-off-by: Ce Gao <gaoce@caicloud.io>


gaocegege:
I migrated the golint check from Kubernetes, and it allows some packages in .golint_failures fails the check.

Another option is gometalinter, and kubeflow uses this way. But I found that kube-arbitrator follows the k8s style gofmt check then I think gometalinter is not our first choice.

PTAL

k82cn:
overall, we follow k8s style for tools, as I'd like to get it merge to k8s/upstream in future :).

k82cn (approved):
lgtm

PR #169, opened 2018-02-16
Make minor fixes to cache.go

kkaneda:
Fix errors reported by "go vet" and "errcheck". The error from "go vet" was following:

```
batchd/cache/cache.go:677: String passes lock by value: cache.SchedulerCache
```


pkg/batchd/cache/cache.go
```diff
@@ -673,7 +674,7 @@ func (sc *SchedulerCache) Snapshot() *CacheSnapshot {
 	return snapshot
 }
 
-func (sc SchedulerCache) String() string {
+func (sc *SchedulerCache) String() string {
```
k82cn, line 677: That should not be a pointer :)

KenjiKanedaNVIDIA: Thanks for the review , @k82cn ! Hmm, if this is not a pointer receiver, why do we call `sc.Mutex.Lock` inside `String`? 

https://medium.com/golangspec/detect-locks-passed-by-value-in-go-efb4ac9a3f2b

kkaneda: Thanks for the review , @k82cn ! Hmm, if this is not a pointer receiver, why do we call sc.Mutex.Lock inside String?

https://medium.com/golangspec/detect-locks-passed-by-value-in-go-efb4ac9a3f2b

k82cn: Good point ! LGTM :)

k82cn (approved):

k82cn:
@jinzhejz , do you know where do we log cache for debug?

davidopp:
Hi @kkaneda !! Great to see you're interested in the kube-arbitrator!

kkaneda:
Hi @davidopp ! Yes, was curious. :) Haven't yet looked very closely, but looks very promising!

k82cn:
@kkaneda , thanks very much for the patch, I'd like to get it merged :)

KenjiKanedaNVIDIA:
thank you, @k82cn !

kkaneda:
thank you, @k82cn !

Issue #164, opened 2018-02-07
Add node selector to identify a sub-cluster for batch job


k82cn:
Currently, we cached all node in `kube-bached` and assign node to pod accordingly. But for batch job, the conflict is so heavy :(. We'll use 'static' zone for batch job before finding a "good" way to work together with kube-scheduler.

List some tasks (including but not limited to)
- [x] Filter pods with scheduler name for `kube-batchd` (#170)
- [x] Filter nodes with labels for `kube-batchd` (#203)
- [ ] Filter nodes with NodeAffinity for `kube-batchd`

sedflix:
@k82cn No one is working on it, right?

@k82cn @jinzhejz 
Is this the requirement? "Nodes, having a particular label X, will be considered by kube-batchd"

k82cn:
The feature was done; close this.

PR #170, opened 2018-02-21
RFC: Add a new option --scheduler-name for ignoring pods which should be h…

mitake:
…andled by the default scheduler

This commit adds a new option --scheduler-name to
kube-batchd. kube-batchd handles pods which have the name specified
with the option in its v1.Pod.Spec.SchedulerName. The motivation is
separating pods which should be controlled by kube-batchd from other
pods which can be handled by the default scheduler.

I'm still not fully sure this change is valuable for kube-arbitrator's design. If I can have comments, I'm really glad.

### DIFF
--- a/pkg/batchd/cache/cache.go
+++ b/pkg/batchd/cache/cache.go
@@ -18,6 +18,7 @@ package cache
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/golang/glog"
@@ -38,8 +39,8 @@ import (
 )
 
 // New returns a Cache implementation.
-func New(config *rest.Config) Cache {
-	return newSchedulerCache(config)
+func New(config *rest.Config, schedulerName string) Cache {
+	return newSchedulerCache(config, schedulerName)
 }
 
 type SchedulerCache struct {
@@ -58,7 +59,7 @@ type SchedulerCache struct {
 	Pdbs   map[string]*PdbInfo
 }
 
-func newSchedulerCache(config *rest.Config) *SchedulerCache {
+func newSchedulerCache(config *rest.Config, schedulerName string) *SchedulerCache {
 	sc := &SchedulerCache{
 		assumedPodStates: make(map[string]*podState),
 		Nodes:            make(map[string]*NodeInfo),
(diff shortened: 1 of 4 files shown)


k82cn:
xref #164

pkg/batchd/cache/cache.go
```diff
@@ -88,6 +89,9 @@ func newSchedulerCache(config *rest.Config) *SchedulerCache {
 			FilterFunc: func(obj interface{}) bool {
 				switch t := obj.(type) {
 				case *v1.Pod:
+					if strings.Compare(obj.(*v1.Pod).Spec.SchedulerName, schedulerName) != 0 {
```
jinzhejz: Only filter Pods with the same scheduler name may be not enough. It may cause `kube-batchd` schedule pods to a node overload.

Such as some pods from another scheduler are running on a node, however, the node is idle for `kube-batchd` due to `kube-batchd` doesn't get those running pods info. When `kube-batchd` try to schedule pods on that node, it may fail due to lack of resources.

k82cn: +1

We should watch 'pending & schedulerName' or 'running' pods.

mitake: @jinzhejz @k82cn thanks for your review. I'll update the condition. How about 'pending & schedulerName' or 'nonTerminatedPod()'?

pkg/batchd/cache/cache.go
```diff
@@ -88,6 +89,10 @@ func newSchedulerCache(config *rest.Config) *SchedulerCache {
 			FilterFunc: func(obj interface{}) bool {
 				switch t := obj.(type) {
 				case *v1.Pod:
+					pod := obj.(*v1.Pod)
+					if strings.Compare(pod.Spec.SchedulerName, schedulerName) == 0 && pod.Status.Phase == v1.PodPending {
```
jinzhejz: it also adds pending pods from other scheduler(`nonTerminatedPod()` will return true)

mitake: ah I see, will fix it soon, thanks

mitake:
@jinzhejz updated the condition, could you take a look?

jinzhejz:
@mitake there is build error, could you take a look

```
# github.com/kubernetes-incubator/kube-arbitrator/pkg/batchd/cache
pkg/batchd/cache/cache.go:90:17: t declared and not used
make: *** [kube-arbitrator] Error 2
```

And other looks good to me.

jinzhejz (approved):
lgtm

geekSiddharth:
@k82cn As per this PR, shouldn't --scheduler-name included in [kube-batchd.yaml](https://github.com/kubernetes-incubator/kube-arbitrator/blob/master/doc/usage/batchd_tutorial.md#2-define-a-kubernetes-deployment-for-kube-batchd)

PR #181, opened 2018-03-16
Support nvidia.com/gpu as a resource managed by kube-…

mitake:
…batchd

This commit lets kube-batchd manage GPU resources
(nvidia.com/gpu) and allocate them for pods like other
resources (CPU and memory).

/cc @YujiOshima

### DIFF
--- a/pkg/batchd/cache/resource_info.go
+++ b/pkg/batchd/cache/resource_info.go
@@ -26,12 +26,19 @@ import (
 type Resource struct {
 	MilliCPU float64
 	Memory   float64
+	GPU      int64
 }
+
+const (
+	// need to follow https://github.com/NVIDIA/k8s-device-plugin/blob/66a35b71ac4b5cbfb04714678b548bd77e5ba719/server.go#L20
+	GPUResourceName = "nvidia.com/gpu"
+)
 
 func EmptyResource() *Resource {
 	return &Resource{
 		MilliCPU: 0,
 		Memory:   0,
+		GPU:      0,
 	}
 }
 
@@ -39,6 +46,7 @@ func (r *Resource) Clone() *Resource {
 	clone := &Resource{
 		MilliCPU: r.MilliCPU,
 		Memory:   r.Memory,
+		GPU:      r.GPU,
 	}
 	return clone
 }
@@ -54,13 +62,16 @@ func NewResource(rl v1.ResourceList) *Resource {
 			r.MilliCPU += float64(rQuant.MilliValue())
 		case v1.ResourceMemory:
 			r.Memory += float64(rQuant.Value())
+		case GPUResourceName:
+			q, _ := rQuant.AsInt64()
+			r.GPU += q
 		}
 	}
 	return r
 }
 
 func (r *Resource) IsEmpty() bool {
-	return r.MilliCPU < minMilliCPU && r.Memory < minMemory
+	return r.MilliCPU < minMilliCPU && r.Memory < minMemory && r.GPU == 0
 }
 
 func (r *Resource) IsZero(rn v1.ResourceName) bool {
@@ -69,6 +80,8 @@ func (r *Resource) IsZero(rn v1.ResourceName) bool {
 		return r.MilliCPU < minMilliCPU
 	case v1.ResourceMemory:
 		return r.Memory < minMemory
+	case GPUResourceName:
+		return r.GPU == 0
 	default:
 		panic("unknown resource")
 	}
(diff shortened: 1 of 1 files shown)


pkg/batchd/cache/resource_info.go
```diff
@@ -26,19 +26,26 @@ import (
 type Resource struct {
 	MilliCPU float64
 	Memory   float64
+	GPU      int64
 }
 
+const (
+	GPUResourceName = "alpha.kubernetes.io/nvidia-gpu"
```
YujiOshima: GPU resource name has been changed from v1.8 onwards. ( https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/#v18-onwards )
You should use ` nvidia.com/gpu` for v1.8 and v1.9.

geekSiddharth:
[quotalloc/cache/quotaallocator_info.go](https://github.com/kubernetes-incubator/kube-arbitrator/blob/f152c8c6233c41f00e454c213504e96efc5b46cc/pkg/quotalloc/cache/quotaallocator_info.go#L35) has a function `CompareResources`. It may require changes.

pkg/batchd/cache/resource_info.go
```diff
@@ -54,13 +61,16 @@ func NewResource(rl v1.ResourceList) *Resource {
...
+			q, _ := rQuant.AsInt64()
+			r.GPU += q
 		}
 	}
 	return r
 }
 
 func (r *Resource) IsEmpty() bool {
-	return r.MilliCPU < minMilliCPU && r.Memory < minMemory
+	return r.MilliCPU < minMilliCPU && r.Memory < minMemory && r.GPU == 0
```
sedflix, line 74: What if GPU is not present on any node in the cluster?

sedflix:
If this PR is alright, I could do the the same for quota-alloc.

k82cn:
@geekSiddharth , please hold the enhancement to kube-quotalloc; we'd like to make kube-batchd ready firstly :)

pkg/batchd/cache/resource_info.go
mitake: I found a useful const in the k8s api package: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4084
Probably using it would be the most straightforward way for the purpose?

YujiOshima: After v1.8, k8s delegated the management of GPU to `Nvidia device plugin`.
So the resource name is defined in `Nvidia device plugin`. https://github.com/NVIDIA/k8s-device-plugin/blob/66a35b71ac4b5cbfb04714678b548bd77e5ba719/server.go#L20

mitake: Unfortunately the const seems to be private. We need to have our own local copy in kube-batchd...

pkg/batchd/cache/resource_info.go
mitake: Pods won't be bond forever.

mitake:
@k82cn I tested this PR with the test command of tf-operator (added in this PR: https://github.com/kubeflow/tf-operator/pull/509), and it seems to be working well for now. Could you take a look?

pkg/batchd/cache/resource_info.go
sedflix: Is that the desired behavior? Will all pod require GPU?

mitake: > Is that the desired behavior? 

Probably yes.

> Will all pod require GPU?

It depends on workloads.

pkg/batchd/cache/resource_info.go
```diff
@@ -85,23 +99,25 @@ func (r *Resource) Sub(rr *Resource) *Resource {
...
 func (r *Resource) Less(rr *Resource) bool {
-	return r.MilliCPU < rr.MilliCPU && r.Memory < rr.Memory
+	return r.MilliCPU < rr.MilliCPU && r.Memory < rr.Memory && r.GPU < rr.GPU
 }
 
 func (r *Resource) LessEqual(rr *Resource) bool {
 	return (r.MilliCPU < rr.MilliCPU || math.Abs(rr.MilliCPU-r.MilliCPU) < 0.01) &&
-		(r.Memory < rr.Memory || math.Abs(rr.Memory-r.Memory) < 1)
+		(r.Memory < rr.Memory || math.Abs(rr.Memory-r.Memory) < 1) &&
+		(r.GPU < rr.GPU || rr.GPU == 0)
```
jinzhejz: @mitake I think it should be` (r.GPU < rr.GPU || r.GPU == rr.GPU)` :)

mitake: You mean `r.GPU <= rr.GPU`?

jinzhejz: yes

k82cn:
@mitake, thanks for your contribution; it'll be better to have a UT for GPU in `drf_test.go`.

mitake:
@k82cn thanks, I'll add the UT later

PR #183, opened 2018-03-19
remove TODO: use /eviction instead of delete to avoid race-condition

rc-zhang:
cc @gaocegege


rc-zhang:
CLA signed

k82cn:
@rc-zhang , thanks very much for your contribution; but I think the `Evict` is similar to `Delete` right now. The pod was deleted and re-created by controller. But what we want to do is NOT to delete (just terminate the pods) the pods in api to avoid race condition.

PR #189, opened 2018-03-22
kube-batchd tutorial packaging and deployment fix

sedflix:
A list of changes made:

- Removed usage config file option: Since kube-batchd is being deployed inside the cluster in this example, we don't need it. Moreover, it creates a series of problems if we are using minikube.
- Added a step to push the image to a registry
- Added `-scheduler-name=kube-batchd` option as per #170


jinzhejz (approved):
lgtm

PR #203, opened 2018-04-26
support NodeSelector for PodSet

jinzhejz:
Fix the second task in issue #164 

The filter is for PodSet level, assume that all pods in same PodSet have the same NodeSelector

### DIFF
--- a/pkg/batchd/policy/drf/drf.go
+++ b/pkg/batchd/policy/drf/drf.go
@@ -61,10 +61,16 @@ func (drf *drfScheduler) Allocate(queues []*cache.QueueInfo, nodes []*cache.Node
 	// assign MinAvailable of each podSet first by chronologically
 	sort.Sort(dq)
 	pq := util.NewPriorityQueue()
+	matchNodesForPodSet := make(map[string][]*cache.NodeInfo)
 	for _, q := range dq {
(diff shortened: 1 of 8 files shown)


pkg/batchd/policy/drf/drf.go
```diff
@@ -64,7 +64,12 @@ func (drf *drfScheduler) Allocate(queues []*cache.QueueInfo, nodes []*cache.Node
 	for _, q := range dq {
 		psi := q.Value.(*podSetInfo)
 
-		assigned := drf.assignMinimalPods(psi.insufficientMinAvailable(), psi, nodes)
+		// fetch the nodes that match PodSet NodeSelector and NodeAffinity
+		// and store it for following DRF assignment
+		matchNodes := fetchMatchNodeForPodSet(psi, nodes)
+		psi.matchNodes = append(psi.matchNodes, matchNodes...)
```
k82cn: why add a `matchNodes` in psi?

jinzhejz: just store it for DRF assignment after minimal assignment, I will remove it from `psi` and use a map in `drf.go`

k82cn (approved):

PR #242, opened 2018-06-22
Example of enhanced queue job controller

scostache:
This PR proposes a new controller for kube-arbitrator, capable of managing more complex requirements, from ML workloads. The PR adds a new definition of the QJ, called XQJ, as well as a controller that manages resources such as pods, services, etc. The PR also adds the extended XQJ controller, that can be easily modified to support different QJ resources.


cmd/kar-controllers/app/server.go
```diff
@@ -41,8 +41,11 @@ func Run(opt *options.ServerOption) error {
 
 	neverStop := make(chan struct{})
 
-	queuejobctrl := queuejob.NewQueueJobController(config)
-	queuejobctrl.Run(neverStop)
+	//queuejobctrl := queuejob.NewQueueJobController(config)
+	//queuejobctrl.Run(neverStop)
```
k82cn, line 45: We need to keep QueueJob controller here.

k82cn:
This PR is so huge, please help to separate it into some smaller PRs, e.g. v1alpha1, client and controller.

cmd/kar-controllers/app/server.go
dmatch01: @scostache Could you comment on the impl. here? ^^^

PR #252, opened 2018-06-27
template for extended QueueJob definition to support ML workloads

scostache:
Part 1 of the previous PR for an enhanced queue job controller. The current PR adds an extended template definition for the QueueJob, called XQueueJob.


k82cn:
Please run `make generate-code` to generate related codes.

PR #274, opened 2018-07-06
Xqueuejob client code part 1

dmatch01:
Part 1 only includes new files.  This is the first part of a xlarge PR for enhanced queuejob.


scostache:
How is this different from my previous PR: 253 ?

dmatch01:
@scostache This is a subset of 253.  I was starting to break up 253 such that it would not be in XXL category.  This was in response to @k82cn request to make the PRs smaller.  

@k82cn Is this PR size workable for you?  Please advise.

deployment/Dockerfile.queue-ctrl
```diff
@@ -0,0 +1,3 @@
+From ubuntu:18.04
```
k82cn, line 1: why we need this dockerfile?

pkg/controller/xqueuejob/queuejob_controller_ex.go
```diff
@@ -0,0 +1,240 @@
+/*
```
k82cn, line 1: please remove the controller in this pr.

scostache: @k82cn : see PR#253

k82cn: Would you sync up before open PRs ?

Issue #285, opened 2018-07-08
schedulerName specified per QueueJob instead of per Pod Template


scostache:
Currently the scheduler name is specified as part of the pod template. When creating a QueueJob with multiple pod "groups" (pods created from different templates), making the user to specify the scheduler name for each template might be error prone. Can we have it as part of the QueueJob template and then filled-in automatically in each pod template?

k82cn:
That's reasonable to me: if no schedulerName in pod template, set it to QueueJob's; otherwise, keep it as is.

scostache:
PR #293

k82cn:
Fixed by  #293

PR #293, opened 2018-07-09
scheduler name can be specified as part of the QueueJob Spec

scostache:
Referencing issue #285

### DIFF
--- a/pkg/controller/queuejob/utils.go
+++ b/pkg/controller/queuejob/utils.go
@@ -95,8 +95,8 @@ func createQueueJobPod(qj *arbv1.QueueJob, template *corev1.PodTemplateSpec, ix 
 		},
 		Spec: templateCopy.Spec,
 	}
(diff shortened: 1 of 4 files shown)


pkg/controller/queuejob/utils.go
```diff
@@ -95,6 +95,11 @@ func createQueueJobPod(qj *arbv1.QueueJob, template *corev1.PodTemplateSpec, ix
 		},
 		Spec: templateCopy.Spec,
 	}
+	// we replace the scheduler name (if any) specified in the pod definition with the one specified in the QJ template
+	if qj.Spec.SchedulerName != "" {
```
k82cn: It's better to check Pod's scheduler name: if it's empty, replace it with QueueJob's schedule name.

scostache: @k82cn : I've also added a check for the schedulerName in the podTemplate based on your comment.

PR #301, opened 2018-07-13
Creating multi-architecture images using docker manifest tool

damora:
Added platform specific Dockerfiles for ppc64le, amd64 and ubuntu and centos
Modified existing Makefile adding a manifest build section:
1. This section pushes the architecture specific images to ppc64le-kube-arbitrator and amd64-kube-arbitrator repositories
2. Creates the manifest that includes
   damora/ppc64le-kube-arbitrator:ubuntu-0.1
    damora/ppc64le-kube-arbitrator:centos-0.1
   damora/amd64-kube-arbitrator:ubuntu-0.1
3. pushes the manifest to damora/kube-arbitrator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 BIN_DIR=_output/bin
+DOCKER_REPO=damora
 RELEASE_VER=0.1
 
 kube-arbitrator: init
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,4 +1,5 @@
 BIN_DIR=_output/bin
+DOCKER_REPO=damora
```
k82cn, line 2: please use kubearbitrator

deployment/Dockerfile.ppc64le.centos
```diff
@@ -0,0 +1,12 @@
+FROM  ppc64le/centos:7
```
k82cn, line 1: Is it necessary to provide two image?

k82cn:
please sign CNCF-CLA :)

deployment/Dockerfile.ppc64le.centos
damora: you mean an Ubuntu and Centos image? Probably not.

Makefile
k82cn: oh, I create kubsigs for us.

jeefy:
I don't think this will work. 

You have to have an Edge build of Docker and enable experimental-CLI features to be able to use the manifest command. Travis doesn't have this version of Docker by default.

See https://github.com/kubernetes/dashboard/pull/3143/files for an example :)

PR #303, opened 2018-07-16
fix the bug: kube-batchd can not aware of the changes of gpu resources when scale up cluster.

GarnettWang:
fix the issue: [kube-batchd can not aware of the changes of gpu resources when scale up cluster. ](https://github.com/kubernetes-incubator/kube-arbitrator/issues/302)


GarnettWang:
when scale up cluster, AddNode does not contain gpu info, as node register to kube-apiserver 
 prior to nvidia device plugin successfully register gpu deivces. 

So UpdateNode should set the Idle Resouces for gpu to Allocatable gpu resouces.

PR #341, opened 2018-08-29
Add Tusimple

suleisl2000:
Add Tusimple


k82cn:
Thanks very much for your interesting :)

PR #353, opened 2018-09-06
fix deployment file invalid error

ChanYiLin:
In the original version, this error happened when using Helm to deploy kube-arbitrator

```
spec.selector: Required value
spec.template.metadata.labels: Invalid value: map[string]string{...} selector does not match template labels
```

This PR add selector in spec of Deployment to fix the invalid error.


ChanYiLin:
I signed it

ChanYiLin:
I signed it

k82cn:
Thanks very much for the PR :)

Issue #364, opened 2018-09-13
Support PDB and add e2e test for it


k82cn:
**FOUND IN VERSION** (Details about the version, e.g. 0.7. Please list the other versions which this bug exist.)
`v0.2`

**PLATFORMS e.g. 'uname -a'**:
[ `Darwin k82cn 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64` ]

**COMMANDS OR DAEMONS**: -- list all related components
 kube-batchd 

**DESCRIPTION**: -- symptom of the problem a customer would see

We used to support PDB as scheduling specific of Gang-Scheduling/Coscheduling. But when we
enhance to support PodGroup, it seems broken ( https://github.com/kubeflow/tf-operator/issues/743#issuecomment-419869576 ). We need to fix it and add a e2e for it.

The PDB will be removed/deprecated until PodGroup merged in upstream.

**IMPACT**: -- impact of problem in customer env (best/worse case scenarios)

 `tf-operator` can not work with `kube-batchd:v0.2` 

**Tasks**:

- [X] apply enabled namespace as Queue to PDB #365
- [ ] e2e test for PDB

ChanYiLin:
I might be a little bit busy these two weeks, but I can add e2e test before September.

k82cn:
>  but I can add e2e test before September.

Great, thanks :)

PR #365, opened 2018-09-13
apply enabled namespace as Queue to PDB

ChanYiLin:
This PR apply "enabled namespace as Queue" to PDB to keep consistency with podgroup
or PDB would not work in the integration with tf-operator.


ChanYiLin:
@k82cn 
I have tested it, this fix makes pdb work with using tf-operator

This fix related to issue #364

k82cn:
Thanks very much, the fix is OK to me. Do you have time to add e2e for it?

PR #367, opened 2018-09-13
Tweak the doc.

tizhou86:
Tweak the tutorial for users who are not familiar with Kubernetes RBAC.

### DIFF
--- a/doc/usage/tutorial.md
+++ b/doc/usage/tutorial.md
@@ -39,7 +39,7 @@ NAME        	REVISION	UPDATED                 	STATUS  	CHART                	NA
 dozing-otter	1       	Thu Jun 14 18:52:15 2018	DEPLOYED	kube-arbitrator-0.2.0	kube-system
 ```
 
-NOTE: `kube-arbitrator` need to collect cluster information(such as Pod, Node, CRD, etc) for scheduing, so the service account used by the deployment must have permission to access those cluster resources, otherwise, `kube-arbitrator` will fail to startup.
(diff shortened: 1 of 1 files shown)


doc/usage/tutorial.md
```diff
@@ -39,7 +39,7 @@ NAME        	REVISION	UPDATED                 	STATUS  	CHART                	NA
 dozing-otter	1       	Thu Jun 14 18:52:15 2018	DEPLOYED	kube-arbitrator-0.2.0	kube-system
 ```
 
-NOTE: `kube-arbitrator` need to collect cluster information(such as Pod, Node, CRD, etc) for scheduing, so the service account used by the deployment must have permission to access those cluster resources, otherwise, `kube-arbitrator` will fail to startup.
+NOTE: `kube-arbitrator` need to collect cluster information(such as Pod, Node, CRD, etc) for scheduing, so the service account used by the deployment must have permission to access those cluster resources, otherwise, `kube-arbitrator` will fail to startup. For users who are not familiar with Kubernetes RBAC, please copy the example/role.yaml into `$GOPATH/src/github.com/kubernetes-incubator/kube-arbitrator/deployment/kube-arbitrator/templates/` and reinstall arbitrator.
```
k82cn, line 42: If copy, will you be able to delete helm release?

tizhou86: yes, the clusterrolebinding resource also get deleted.

k82cn: great, thanks.

PR #377, opened 2018-09-27
 Add directories for extended queuejob

dmatch01:


dmatch01:
@scostache Can you review and add the /lgtm label once you review and approve?

deployment/kube-queue-ctrl.yaml
```diff
@@ -0,0 +1,40 @@
...
+        component: controller
+        tier: control-plane
+        version: second
+    spec:
+      containers:
+      - command:
+        - /bin/sh
+        - -c
+        - touch /tmp/healthy; /opt/kube-queue-ctrl --scheduler-name kube-batchd; rm -f /tmp/healthy
+        image: kubearbitrator/queue-ctrl:v0.1
```
k82cn: we should not use this repos; it's changed to `kubesigs/kube-batch`

pkg/apis/controller/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
k82cn, line 2: Why introduce those `.gitignore` files?

deployment/kube-queue-ctrl.yaml
k82cn: And where is the dockerfile build this images? we already move them out.

pkg/apis/controller/.gitignore
dmatch01: Per your past recommendation to submit smaller sized PRs to improve merge consume-ability this first PR submitted for your review creates isolated directories to contain the extended queuejob thus the gitignore files are needed.  As we submit future PRs to add to these directories the gitinore files will be adjusted/removed.

deployment/kube-queue-ctrl.yaml
dmatch01: Removed from PR and will make appropriate changes when re-adding to a future PR.

pkg/apis/controller/.gitignore
k82cn: but .gitignore does not make sense to me.

dmatch01: Was you question about the content of the gitignore file?  My understanding to make a directory empty (in the repository) that will checkout after a merge is to create a .gitignore file inside that empty directory that contains those lines.  As we submit future PRs to add to these directories the gitinore files will be adjusted/removed.  If you prefer not to have initial directories empty on the first PR I can add a few of the files that will be within each directory to keep the PR small/med in size and remove the gitignore file.  Please advise.

Issue #409, opened 2018-10-10
All jobs are pending, when some jobs set resources，others not set resources. 


chenyangxueHDU:
I created 4 jobs which have same `group-name`, such as ps-0, ps-1, worker-0, worker-1. Only worker jobs had `resources` tag. 
```yaml
- apiVersion: batch/v1
  kind: Job
  metadata:
    name: cyx2-worker-0
        annotations:
          scheduling.k8s.io/group-name: cyx2
  spec:
   template:
     spec:
       containers:
       -  resources:
            limits:
              nvidia.com/gpu: "1"
            requests:
              cpu: "1"
              memory: 1Gi
- apiVersion: batch/v1
  kind: Job
  metadata:
    name: cyx2-ps-0
        annotations:
          scheduling.k8s.io/group-name: cyx2
  spec:
   template:
     spec:
       containers:
```

I found all jobs are pending. I checked log.
```
I1009 21:42:36.472045   21605 allocate.go:42] Enter Allocate ...
I1009 21:42:36.472224   21605 allocate.go:118] Binding Task <mind-automl/cyx2-worker-0-2mr7q> to node <192.168.47.52>
I1009 21:42:36.472399   21605 allocate.go:118] Binding Task <mind-automl/cyx2-worker-1-hdz8r> to node <192.168.47.52>
I1009 21:42:36.472426   21605 allocate.go:72] Queue <mind-automl> is overused, ignore it.
I1009 21:42:36.472431   21605 allocate.go:155] Leaving Allocate ..
```
I found queue is overuserd and only bind workers, but I did not set queue. I found the key in `kube-batch\pkg\scheduler\plugins\proportion\proportion.go`.
```go
                        // Calculates the deserved of each Queue.
			attr.deserved.Add(remaining.Clone().Multi(float64(attr.weight) / float64(totalWeight)))
			if !attr.deserved.LessEqual(attr.request) {
				attr.deserved = helpers.Min(attr.deserved, attr.request)
				meet[attr.queueID] = struct{}{}
			}
```
These mean I can use resources what I request. Then I set ps jobs resources, all jobs are running. I think it should be reminded set resources in tutorial.
```go
			if !attr.deserved.LessEqual(attr.request) {
				attr.deserved = helpers.Min(attr.deserved, attr.request)
				meet[attr.queueID] = struct{}{}
			}
```
Or we can change this.

k82cn:
We should fix it :)

[...]

PR #416, opened 2018-10-11
fix BestEffort for overused

chenyangxueHDU:
fix ths issue: [All jobs are pending, when some jobs set resources，others not set resources.](https://github.com/kubernetes-sigs/kube-batch/issues/409)

### DIFF
--- a/pkg/scheduler/plugins/priority/priority.go
+++ b/pkg/scheduler/plugins/priority/priority.go
@@ -20,6 +20,8 @@ import (
 	"github.com/golang/glog"
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api"
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/framework"
+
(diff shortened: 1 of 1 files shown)


k82cn:
Please fix this in master, and then cherry pick to release-0.2. Can we add e2e test for it?

pkg/scheduler/plugins/priority/priority.go
```diff
@@ -37,6 +39,10 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 		lv := l.(*api.TaskInfo)
 		rv := r.(*api.TaskInfo)
 
+		if res := compareQoS(lv.Pod, rv.Pod); res != 0 {
```
k82cn, line 42: Please add a TODO here, that we're going to enhance it later by other option.

chenyangxueHDU:
> Please fix this in master, and then cherry pick to release-0.2. Can we add e2e test for it?

OK

Issue #422, opened 2018-10-13
Log message wrong


chenyangxueHDU:
Line 84~85 in kube-batch\pkg\scheduler\plugins\predicates\predicates.go

```go
			return fmt.Errorf("task <%s/%s> does not tolerate node <%s> taints",
				node.Name, task.Namespace, task.Name)
```
it should be

```go
			return fmt.Errorf("task <%s/%s> does not tolerate node <%s> taints",
				task.Namespace, task.Name, node.Name)
```

PR #423, opened 2018-10-13
fix err log

chenyangxueHDU:
fix the issue: [Log message wrong #422](https://github.com/kubernetes-sigs/kube-batch/issues/422)


k82cn (commented):
/lgtm
/approve

PR #440, opened 2018-10-15
Updated node info if taints and lables changed.

### LINKED ISSUE #424: kube-batchd can not aware of the changes node except resources
1. cordon node1
1. start kube-batch
1. create a job on node1, then pods are pending for taints
1. uncordon node1
1. pods still are pending for taints

I check the code in `kube-batch\pkg\scheduler\cache\event_handlers.go`
```go
func isNodeInfoUpdated(oldNode, newNode *v1.Node) bool {
	return !reflect.DeepEqual(oldNode.Status.Allocatable, newNode.Status.Allocatable)
}

// Assumes that lock is already acquired.
func (sc *SchedulerCache) updateNode(oldNode, newNode *v1.Node) error {
	if sc.Nodes[newNode.Name] != nil {
		if isNodeInfoUpdated(oldNode, newNode) {
			sc.Nodes[newNode.Name].SetNode(newNode)
		}
		return nil
	}

	return fmt.Errorf("node <%s> does not exist", newNode.Name)
}
```
I think we can not only care about resources, and ignore other changes which the scheduler care about, such as taints.

k82cn:

fixes: #424

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -241,7 +241,9 @@ func (sc *SchedulerCache) addNode(node *v1.Node) error {
 }
 
 func isNodeInfoUpdated(oldNode, newNode *v1.Node) bool {
-	return !reflect.DeepEqual(oldNode.Status.Allocatable, newNode.Status.Allocatable)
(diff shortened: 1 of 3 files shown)


k82cn:
will add an e2e for it later.

PR #465, opened 2018-11-05
Enable PDB based gang scheduling with discrete queues.

adam-marek:
Currently it's not possible to schedule PDB based gangs when the `enable-namespace-as-queue` option is disabled. This PR introduces an option to set the name of the queue to be used with PDB based jobs via the command line (if not specified the name of the namespace is still used).
Fixes #

**Release note**:

```release-note
It's now possible to scheduled PDB based jobs while the `enable-namespace-as-queue` option is set to `false`. The new `pdb-queue` command line option allows to manually select the queue to be used for PDB based jobs in such cases.
```

### DIFF
--- a/cmd/kube-batch/app/options/options.go
+++ b/cmd/kube-batch/app/options/options.go
@@ -33,6 +33,18 @@ type ServerOption struct {
 	NamespaceAsQueue     bool
 	EnableLeaderElection bool
 	LockObjectNamespace  string
+	PdbQueue             string
+}
+
+var (
+	opts *ServerOption
+)
+
+func Options() *ServerOption {
+	if opts == nil {
+		opts = &ServerOption{}
+	}
+	return opts
 }
 
 // NewServerOption creates a new CMServer with a default config.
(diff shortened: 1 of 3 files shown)


cmd/kube-batch/app/options/options.go
```diff
@@ -49,6 +61,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.SchedulerName, "scheduler-name", "kube-batch", "kube-batch will handle pods with the scheduler-name")
 	fs.StringVar(&s.SchedulerConf, "scheduler-conf", "", "The namespace and name of ConfigMap for scheduler configuration")
 	fs.StringVar(&s.SchedulePeriod, "schedule-period", "1s", "The period between each scheduling cycle")
+	fs.StringVar(&s.PdbQueue, "pdb-queue", "", "The name of the Queue object to be used with PDBs instead of their namespace name")
```
k82cn: Comparing to add a parameter, prefer to add an annotation to PDB for `queue`.

k82cn: for example `scheduling.k8s.io/queue-name`

k82cn: with annotation, you can create more queues for different PDBs :)

adam-marek: The specific use case I'm looking at is co-scheduling tf-operator jobs with other multi-node workloads (PodGroup/Queue based). Fully agree that annotations are more flexible, however if we go that route then tf-operator needs to be extended to add said annotations to its PDBs (either directly or via a mutating admission controller) which is problematic for existing deployments.

So we could consider this PR as a stop gap for environments where PDBs cannot be easily equipped with additional annotations (for whatever reason). 

Separately the longer term strategy should be decided: either introduce the annotations and extend the tf-operator's PDBs to include them or switch the latter to PodGroups entirely (since with annotations we would introduce kube-batch specific concepts anyway)?

k82cn: oh, for `tf-operator` case, I'd like to switch it to `PodGroups`; but we need to align with tf-operator team and sig-scheduling for it. How about open an issue in `kubeflow/tf-operator` for it firstly?

k82cn: btw, you did not use multiple queues, right?

k82cn: if we switch to `PodGroup`, we still need to provide default queue name for your case (similar to this PR); otherwise, tf-operator need to have a field for `Queue`.

`Queue` is an experimental feature right now, which means you can have a try, but other community should not dependent on it.

Because of above reasons, maybe your PR is better as a temp solution; WDYT?

k82cn: BTW, prefer `DefaultQueue` instead of `PdbQueue`; so when we remove PDB, it also works for `PodGroup`.

adam-marek: makes sense, will amend

PR #468, opened 2018-11-07
Detailed 'unschedulable' events

adam-marek:
kube-batch currently provides very little information about the resource constraints that keep specific pods from being scheduled. The 'unschedulable' events only state that "there are insufficient resources" for a job without giving any details which makes it hard to troubleshoot "stuck" jobs without resorting to kube-batch logifile analysis.

This PR attempts to expose some of these details to the user and bring the "unschedulable" events more in line with what is available in the stock scheduler. 
It does so by introducing the following changes:
- 'unschedulable' events are reported both for the PodGroup/PDB and the ingredient pods
- 'unschedulable' events provide information about the number of nodes available for scheduling and which resource unavailability caused scheduling to fail for the job
- 'unschedulanle' events include information about the number of gang members that couldn't be successfully allocated existing resources. 

 
Fixes #

**Release note**:

```release-note
NONE
```


pkg/scheduler/api/resource_info.go
```diff
@@ -109,6 +109,25 @@ func (r *Resource) Sub(rr *Resource) *Resource {
 		r, rr))
 }
 
+//Computes the delta between a resource oject representing available
+//resources an operand representing resources being requested.  Any
+//field that is less than 0 after the operation represents an
+//insufficient resource.
+func (r *Resource) FitDelta(rr *Resource) *Resource {
```
k82cn, line 116: this should be a helper function in scheduler instead of a func of resources. `FitDelta` is more about scheduling part.

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -144,8 +145,10 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
 
 func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 	for _, job := range ssn.Jobs {
-		if len(job.TaskStatusIndex[api.Allocated]) != 0 {
-			ssn.Backoff(job, arbcorev1.UnschedulableEvent, "not enough resource for job")
+		if len(job.TaskStatusIndex[api.Pending]) != 0 {
```
k82cn, line 148: When `Gang-scheduling` did not meet, we need to record an event at job level; if `gang-scheduling` does not meet, the task is allocated but not bind to the host.

pkg/scheduler/cache/cache.go
```diff
@@ -550,5 +550,11 @@ func (sc *SchedulerCache) Backoff(job *arbapi.JobInfo, event arbcorev1.Event, re
 		return fmt.Errorf("no scheduling specification for job")
 	}
 
+	for _, tasks := range job.TaskStatusIndex {
+		for _, t := range tasks {
+			sc.recorder.Eventf(t.Pod, v1.EventTypeWarning, string(event), reason)
```
k82cn, line 555: why do we record event for all task/pods ?

pkg/scheduler/plugins/gang/gang.go
k82cn: btw, we need to support `totalTasksNum > minMemberNum` case for elastic jobs, e.g. Spark, in future.

pkg/scheduler/api/resource_info.go
adam-marek: makes sense

pkg/scheduler/plugins/gang/gang.go
adam-marek: > When `Gang-scheduling` did not meet, we need to record an event at job level; if `gang-scheduling` does not meet, the task is allocated but not bind to the host.

Not sure what you mean here. There may be gang jobs that failed scheduling with no `Allocated` tasks (all `Pending`), right ?

adam-marek: > btw, we need to support `totalTasksNum > minMemberNum` case for elastic jobs, e.g. Spark, in future.

When calculating the total number of task, the size of JobInfo->Tasks is used, which should cover the case you mention, correct ?

pkg/scheduler/cache/cache.go
adam-marek: This is done in order to match the stock scheduler behaviour which kind of sets the norm in K8s world in terms of users' expectations. No events recorded for Pods which "hang" in `Pending` state for a long time might seem weird ... 
Also this aids in quicker troubleshooting, as one does not need to track down the correct higher level entity to find out why a pod appears to be stuck.

pkg/scheduler/api/resource_info.go
adam-marek: on a second thought though, it is generic in the sense that it computes the delta or deficit of resource regardless of the context ... also it depends on a number of intimate details of resources (such as minMilliCPU, etc.) which would need to be exported if we put that in a different package ..
thoughts ?

k82cn: what's different with `Sub`?

adam-marek: `Sub` does not allow for negative results (the left operand must be greater or equal to the right operand).

k82cn:
/lgtm

LGTM overall, let's get it merged :)

Issue #521, opened 2018-12-28
Add phase/conditions into PodGroup.Status


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

Currently, kube-batch only raised `Unschedulable` event for its status; it's better to include phase and conditions in `PodGroup.Status` to give more detail.

**Tasks**

- [x] Design Doc (https://github.com/kubernetes-sigs/kube-batch/pull/533)
- [x] Add phase/condition into `PodGroup.Status` CRD (https://github.com/kubernetes-sigs/kube-batch/pull/525)
- [x] Updated phase/condition by `kube-batch` (https://github.com/kubernetes-sigs/kube-batch/pull/560)

Zyqsempai:
@k82cn I would like to work on this one, but can you please provide more details.
As far as I can see, currently we have that type of status for `PodGroup`:
https://github.com/kubernetes-sigs/kube-batch/blob/master/pkg/apis/scheduling/v1alpha1/types.go#L50
You want to extend `PodGroupStatus` struct and add few fields for additional info?

k82cn:
> You want to extend PodGroupStatus struct and add few fields for additional info?

Yes, we need to add `PodGroupPhase` and `Conditions` at least; so the operator/controller knows where it is in its lifecycle. We can refer to PodStatus :)

Zyqsempai:
Got it, I will take care of it!

Zyqsempai:
@k82cn Hey, so I opened PR, but I still have some questions, I am not sure how and where we will fill up those new fields, also, I created them based on same Pod.Status fields, not sure that every value is acceptable for PodGroup.

k82cn:
>I am not sure how and where we will fill up those new fields

kube batch will update after each scheduling cycle.

Zyqsempai:
@k82cn , Great, will be happy to work on it, but what about original PR, should we merge it, or wait for related PR based on your design doc?

k82cn:
> but what about original PR, should we merge it, 

For your PR, I think that's a good start; I'll also add some comments according to the design doc. So I'd suggest to update your PR when design doc is ready. I'll try to get doc ready this week and ask you and @Jeffwan to review

Jeffwan:
@k82cn I think for short term, I will only update podCondition for those pods that fails in predicate action. If that case, cluster autoscaler will scale up right count of new nodes. It would be great to see your doc and work on long term support.

[...]

PR #533, opened 2019-01-03
The design doc of PodGroup Phase in Status.

k82cn:

Part of #521

**Release note**:
```release-note
None
```

### DIFF
--- a/doc/design/podgroup-status.md
+++ b/doc/design/podgroup-status.md
@@ -1,0 +1,121 @@
+# PodGroup Status Enhancement
+
+@k82cn; Jan 2, 2019
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Motivation](#motivation)
+* [Function Detail](#function-detail)
+* [Feature Interaction](#feature-interaction)
+   * [Cluster AutoScale](#cluster-autoscale)
+   * [Operators/Controllers](#operatorscontrollers)
+* [Reference](#reference)
+
+## Motivation
+
+In [Coscheduling v1alph1](https://github.com/kubernetes/enhancements/pull/639) design, `PodGroup`'s status
+only includes counters of related pods which is not enough for `PodGroup` lifecycle management. More information
+about PodGroup's status will be introduced in this design doc for lifecycle management, e.g. `PodGroupPhase`.
+
+## Function Detail
+
+To include more information for PodGroup current status/phase, the following types are introduced:
+
+```go
+// PodGroupPhase is the phase of a pod group at the current time.
+type PodGroupPhase string
+
+// These are the valid phase of podGroups.
+const (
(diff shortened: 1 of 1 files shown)


doc/design/podgroup-status.md
```diff
@@ -0,0 +1,119 @@
...
+### Cluster AutoScale
+
+[Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) is a tool that
+automatically adjusts the size of the Kubernetes cluster when one of the following conditions is true:
+
+* there are pods that failed to run in the cluster due to insufficient resources,
+* there are nodes in the cluster that have been underutilized for an extended period of time and their pods can be placed on other existing nodes.
+
+For first scenario, the Cluster-Autoscaler need to check PodGroup's `status.State` to know whether there're
+sufficient resources to run the whole group of pods.
```
MaciekPytel: This is not how CA works. It doesn't compute required resources at all. 

Very roughly the way it works is:
0. (Import scheduler predicates from k/k repo)
1. If there are pending pods (pods wit PodScheduled condition set to false with reason given as unschedulable):
1.1 Create an in-memory node object that represents a new node that would result from scale-up.
1.2 Run imported scheduler predicates on pending pods to "schedule" them on in-memory node.
1.3 If some pods remain pending goto 1.1.

This way autoscaler automatically takes into account all scheduling constraints: resources, taints, affinity, storage, etc. Unfortunately it also means it's fundamentally incompatible with any form of custom scheduling. If you want CA to work with kube-batch you need to either encapsulate kube-batch logic in a form of predicates or else duplicate it in CA somehow. It certainly won't be a simple fix. 
It would get even worse for things like gang scheduling. You can swap predicates for something different, but ultimately CA expects a black box that takes (pod, node) pair and tells whether the pod could be scheduled on a given node.

k82cn: What will happen if new node can not be scheduled because of customized plugin? Will it continue to create new nodes?

k82cn: I'm thinking whether CA can keep adding new nodes (with some rate & restriction) if `PodGroup` is pending.

MaciekPytel: CA will check what default scheduler would do with the pending pods. If CA thinks the pod can be scheduled on existing nodes it will not add new nodes.

Jeffwan: That would be hard to work on plugins that can be compatible with CA in the future. In other thread, you mentioned  make pod unschedulable is just tip of an iceber. Do you think we can extract some common criterions in CA for other custom scheduler to meet?  For example, podCondition is just one of them.

k82cn: en. I'm thinking to make CA support as out-of-scope for now; as CA need to work together with default scheduler. We can consider to enhance that in later release.

PR #587, opened 2019-02-13
Adding node priority 

### LINKED ISSUE #520: Support prioritize algorithm in kube-batch

/kind feature

**Description**:

There're several requests on this feature, so I'd like to include it in next release :)

xref https://github.com/kubernetes-sigs/kube-batch/issues/516 , https://github.com/kubernetes-sigs/kube-batch/issues/485 , https://github.com/kubernetes-sigs/kube-batch/issues/448

thandayuthapani:
Signed-off-by: thandayuthapani [thandayuthapani.s@huawei.com](mailto:thandayuthapani.s@huawei.com)
This PR includes the implementation for Node Priority Feature in Kube-Batch

Fixes #520 

Have added priority only in allocate action, will add it in preempt action once the implementation for allocate is reviewed.


thandayuthapani:
CLA Signed

pkg/scheduler/plugins/priority/priority.go
```diff
@@ -74,6 +164,82 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 	}
 
 	ssn.AddJobOrderFn(pp.Name(), jobOrderFn)
+
+	priorityFn := func(task *api.TaskInfo, node *api.NodeInfo, nodes []*api.NodeInfo) (int, error) {
+		nodeInfo := cache.NewNodeInfo(node.Pods()...)
```
k82cn: let's move this into a new plugin, named `prioritize`.

pkg/scheduler/util/priorities/image_locality.go
```diff
@@ -0,0 +1,104 @@
+/*
```
k82cn: move this into plugin directory; do not put plugin specific util in this directory.

pkg/scheduler/util/priorities/image_locality.go
```diff
@@ -0,0 +1,104 @@
...
+Copyright 2016 The Kubernetes Authors.
```
k82cn: s/2016/2019/

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -125,34 +127,47 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 					glog.V(3).Infof("Predicates failed for task <%s/%s> on node <%s>: %v",
 						task.Namespace, task.Name, node.Name, err)
 					continue
+				} else {
+					predicateNodes = append(predicateNodes, node)
+				}
+			}
+			for _, node := range predicateNodes {
+				score, err := ssn.PriorityFn(task, node, predicateNodes)
```
k82cn: why we need `predicateNodes`? it's better to use all nodes, there're already some enhancement in upstream.

pkg/scheduler/plugins/priority/priority.go
```diff
@@ -74,6 +164,82 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		nodeInfo.SetNode(node.Node)
+		var score int = 0
+
+		host, err := priority.ImageLocalityPriorityMap(task.Pod, len(nodes), nodeInfo)
```
k82cn: only include "image locality", "LeastRequestedPriorityMap" and "pod affinity" should be enough.

k82cn:
please make CI happy and add e2e test for that.

k82cn:
just go though the code diff, I'll review it in detail later :)

pkg/scheduler/plugins/priority/priority.go
k82cn: and "node affinity".

pkg/scheduler/actions/allocate/allocate.go
thandayuthapani: We need to run priority on predicated nodes right? If we run priority in every nodes and a node which does not satify predicates gets high score, then it will try to Allocate on that node right? So we should run priority on nodes which has already satisfied all the predicates defined right?

k82cn: > We need to run priority on predicated nodes right?

Yes, but why pass all predicateNodes to priority fn?

thandayuthapani: InterPodAffinityPriority function expects set of nodes as input, so we had to send all the predicated nodes to PriorityFn

k82cn: `InterPodAffinityPriority` should use all nodes instead of predicated nodes for topology, as some pods maybe in topology but not in predicated nodes. It seems handled in upstream.

pkg/scheduler/util.go
```diff
@@ -33,6 +33,7 @@ tiers:
 - plugins:
   - name: priority
   - name: gang
+  - name: prioritize
```
k82cn: move this to tier 2.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -132,14 +132,13 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 				}
 			}
 			for _, node := range predicateNodes {
-				score, err := ssn.PriorityFn(task, node, predicateNodes)
+				score, err := ssn.PriorityFn(task, node, ssn.Nodes)
```
k82cn: it's not necessary to pass node from actions; we already has session info in `OpenSession`.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -293,3 +297,26 @@ func (ssn *Session) PredicateFn(task *api.TaskInfo, node *api.NodeInfo) error {
 	}
 	return nil
 }
+
+func (ssn *Session) PriorityFn(task *api.TaskInfo, node *api.NodeInfo) (int, error) {
+	var priorityScore int
+	for _, tier := range ssn.Tiers {
+		for _, plugin := range tier.Plugins {
+			if plugin.PredicateDisabled {
```
k82cn: PriortizedDisable?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -293,3 +297,26 @@ func (ssn *Session) PredicateFn(task *api.TaskInfo, node *api.NodeInfo) error {
...
+			pfn, found := ssn.priorityFns[plugin.Name]
+			if !found {
+				continue
+			}
+			score, err := pfn(task, node)
+			if err != nil {
+				return 0, err
+			} else {
+				priorityScore = score
+				break
```
k82cn: we should get the summary of all plugins as score; and we need to support `weight` later.

pkg/scheduler/plugins/prioritize/priorities/image_locality.go
```diff
@@ -0,0 +1,104 @@
+/*
```
k82cn: why do we need this file?

thandayuthapani: ImageLocalityPriority Function expects priorityMetadata struct as argument, to create that struct we need PriorityMetadataFactory struct, for which we need ServiceLister, ControllerLister, ReplicaSetLister, StatefulSetLister interface to be implemented, but ImageLocalityPriority Function only need totalNumNodes only, so have modified function that it expects only totalNumNodes alone.

k82cn: Do not copy it or modify it; we do not plan effort to keep the compatbility with upstream.

thandayuthapani: Then we will implement our own logic to calculate score for image locality priority, because we will not be able to create PriorityMetadataFactory, without that priorityMetadata struct cannot be also created, so we cannot use image locality priority function provided by kubernetes.

thandayuthapani: How do you want me to handle it?

k82cn: Open an issue in upstream to publish `priorityMetadata`; remove it from this PR right now, and leave a `TODO` there.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -175,4 +189,20 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
+func selectBestNode(nodeScores map[int][]*api.NodeInfo) []*api.NodeInfo {
+	var nodesInorder []*api.NodeInfo
+	var keys []int
+	for key := range nodeScores {
+		keys = append(keys, key)
+	}
+	sort.Ints(keys)
+	for _, key := range keys {
+		nodes := nodeScores[key]
+		for _, node := range nodes {
```
k82cn: `append(nodesInorder, nodes...)`

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -251,3 +267,19 @@ func validateVictims(victims []*api.TaskInfo, resreq *api.Resource) error {
 
 	return nil
 }
+
+func selectBestNode(nodeScores map[int][]*api.NodeInfo) []*api.NodeInfo {
```
k82cn: move to util to avoid duplication.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -293,3 +297,26 @@ func (ssn *Session) PredicateFn(task *api.TaskInfo, node *api.NodeInfo) error {
 	}
 	return nil
 }
+
+func (ssn *Session) PriorityFn(task *api.TaskInfo, node *api.NodeInfo) (int, error) {
+	priorityScore := make(map[string]int)
```
k82cn: why use map?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -293,3 +297,26 @@ func (ssn *Session) PredicateFn(task *api.TaskInfo, node *api.NodeInfo) error {
...
+			}
+			pfn, found := ssn.priorityFns[plugin.Name]
+			if !found {
+				continue
+			}
+			score, err := pfn(task, node)
+			if err != nil {
+				return 0, err
+			} else {
+				priorityScore[plugin.Name] = score
```
k82cn: should be `+=` for summary.

pkg/scheduler/plugins/prioritize/prioritize.go
```diff
@@ -0,0 +1,198 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prioritize
+
+import (
+	"fmt"
```
k82cn, line 20: the import order should be 

```
system

github

k8s

kube-batch
```

pkg/scheduler/plugins/prioritize/prioritize.go
```diff
@@ -0,0 +1,198 @@
...
+		score = score + host.Score
+
+		host, err = priorities.CalculateNodeAffinityPriorityMap(task.Pod, nil, nodeInfo)
+		if err != nil {
+			glog.V(3).Infof("Calculate Node Affinity Priority Failed because of Error: %v", err)
+			return 0, err
+		}
+		score = score + host.Score
+
+		if len(interPodAffinityScore) == 0 {
```
k82cn: why check this? if pod "dispatch" to the node,  the score maybe changed.

k82cn:
please add e2e test for that.

pkg/scheduler/framework/session_plugins.go
thandayuthapani: Once weight is supported, if we are giving different weight to different plugins, it will be easy to calculate score with map.  I will change it to int and return the total sum as score

k82cn: when support weight, we also only return the summary.

test/e2e/prioritize.go
```diff
@@ -0,0 +1,219 @@
...
+	schedulerapi "k8s.io/kubernetes/pkg/scheduler/api"
+)
+
+var _ = Describe("Prioritize E2E Test", func() {
+	It("Node Affinity Test", func() {
+		context := initTestContext()
+		defer cleanupTestContext(context)
+
+		var preferredSchedulingTermSlice []v1.PreferredSchedulingTerm
+		nodeSelectorRequirement := v1.NodeSelectorRequirement{Key: "kubernetes.io/hostname", Operator: v1.NodeSelectorOpIn, Values: []string{"kube-node-2"}}
```
k82cn: get node from node list instead of hardcode; we also need to support different environment, e.g. `./hack/local-up-cluster.sh`

test/e2e/util.go
```diff
@@ -780,6 +781,21 @@ func removeTaintsFromAllNodes(ctx *context, taints []v1.Taint) error {
...
 }
 
+func getAllNodes(ctx *context) []string {
+	nodeNames := make([]string, 0)
+
+	nodes, err := ctx.kubeclient.CoreV1().Nodes().List(metav1.ListOptions{})
+	Expect(err).NotTo(HaveOccurred())
+
+	for _, node := range nodes.Items {
+		if strings.Contains(node.Name, "master") {
```
k82cn: check `node.Spec.Taints` instead of string. And we can not make sure node name contains "master" is the master.

test/e2e/util.go
```diff
@@ -780,6 +781,21 @@ func removeTaintsFromAllNodes(ctx *context, taints []v1.Taint) error {
 	return nil
 }
 
+func getAllNodes(ctx *context) []string {
```
k82cn: we're skipping master node, it's not all nodes.

test/e2e/util.go
```diff
@@ -43,6 +43,7 @@ import (
 	kbv1 "github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
 	kbver "github.com/kubernetes-sigs/kube-batch/pkg/client/clientset/versioned"
 	kbapi "github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api"
+	"strings"
```
k82cn: move this to sys import area.

pkg/scheduler/plugins/prioritize/prioritize.go
```diff
@@ -0,0 +1,197 @@
...
+		var nodeSlice []*v1.Node
+		var interPodAffinityScore schedulerapi.HostPriorityList
+
+		nodeMap, nodeSlice = generateNodeMapAndSlice(ssn.Nodes)
+
+		nodeInfo := cache.NewNodeInfo(node.Pods()...)
+		nodeInfo.SetNode(node.Node)
+		var score = 0
+
+		//TODO: Add ImageLocalityPriority Function once priorityMetadata is published
```
k82cn, line 181: please add issue number here; otherwise, we can not trace the status.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -18,7 +18,6 @@ package allocate
 
 import (
 	"github.com/golang/glog"
-
```
k82cn: do not delete this line

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -125,34 +127,46 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
+			for _, node := range predicateNodes {
+				score, err := ssn.PriorityFn(task, node)
+				if err != nil {
+					glog.V(3).Infof("Error in Calculating Priority for the node:%v", err)
+				} else {
+					nodeScores[score] = append(nodeScores[score], node)
+				}
+			}
+			selectedNodes := util.SelectBestNode(nodeScores)
+			for _, selectedNode := range selectedNodes {
```
k82cn: keep var's name to node to avoid unnecessary diff.

test/e2e/prioritize.go
```diff
@@ -0,0 +1,222 @@
...
+
+		_, pg := createJobEx(context, job)
+		err := waitPodGroupReady(context, pg)
+		Expect(err).NotTo(HaveOccurred())
+
+		pods := getPodOfPodGroup(context, pg)
+		// All Pods Should be Scheduled in same node
+		nodeName := pods[0].Spec.NodeName
+		for _, pod := range pods {
+			Expect(pod.Spec.NodeName).To(Equal(nodeName))
```
k82cn: `rep` means all resources from all worker nodes; they can not be dispatched into the same node. But how this test passed?

thandayuthapani: rep is not all resources from all worker nodes.
```
	for _, node := range nodes.Items {
            ........
            ........
                alloc := kbapi.NewResource(node.Status.Allocatable)
		for slot.LessEqual(alloc) {
			alloc.Sub(slot)
			res++
		}

		if res > 0 {
			return node.Name, res
		}
	}
```
It is number of workload, for particular amount of requested resource that can be scheduled in that node.  If `res` value is greater than zero, then it returns the value for that particular worker node, it does not traverse through all the worker nodes to calculate `res`.  If for that particular workload's requested resource cannot be scheduled then `res` value will be zero, so it checks next node, if `res` is not zero, it does not check other nodes, it will return those values

k82cn: yes, your're right; so that's ok to me :)

k82cn: btw, it seems not stable for this case; would you help to check the CI result?

thandayuthapani: Yeah checking CI failures, will fix it.

test/e2e/predicates.go
```diff
@@ -126,30 +144,26 @@ var _ = Describe("Predicates E2E Test", func() {
...
+		err = waitPodGroupReady(context, pg1)
 		Expect(err).NotTo(HaveOccurred())
 
-		pods := getPodOfPodGroup(context, pg)
-		// All pods should be scheduled to the same node.
-		nodeName := pods[0].Spec.NodeName
-		for _, pod := range pods {
-			Expect(pod.Spec.NodeName).To(Equal(nodeName))
-		}
+		podsWithAffinity := getPodOfPodGroup(context, pg1)
```
k82cn: why do we change that?

thandayuthapani: All the pods are getting unschedulable, so all the pods are in pending state, so it is getting time out while checking `waitPodGroupReady`, so have changed the test case such that, first a task with one pod and label is deployed, and then second task with podAffinity is scheduled, and checking whether both are having same node name, verifying pod affinity predicate is working.

thandayuthapani: screenshot from 2019-02-20 16-18-26
This is resource of DIND-nodes, that is being created in travis, one node always has difference in resources, so if is it trying to schedule pods in that node, for the `rep` value of other nodes, then one pod goes unschedulable, so `waitPodGroupReady` returns error, so test case fails, to avoid that case, this case where one pod is created with label and next pod with `podAffinity` and check is done whether both pods are same scheduled in same node or not, that verifies our pod affinity predicate and avoid unschedulable scenario.

k82cn: if so, just  change oneCpu to halfCpu when creating the Job should be enough.

thandayuthapani: In InterPodAffinity Priority function, they try to get the node of existingPod from it's Pod object, 
```
processPod := func(existingPod *v1.Pod) error {
		existingPodNode, err := ipa.info.GetNodeInfo(existingPod.Spec.NodeName)
                ........
                ........
    }
```
But in gang scheduling, we bind task only when the number of tasks is greater or equal to the MinAvailable number.
```
        occupied := readyTaskNum(job)
	return occupied >= job.MinAvailable
```
So if our job has two tasks, and first get allotted to a node and the cache(schedulercache) info in our scheduler gets reflected with the new pod(v1.Pod) Object, but that pod object will not have nodeName(pod.Spec.NodeName), so when second task comes and checks for InterPodAffinity score, it checks for all the pods from the cache(schedulercache.NodeInfo), and tries to find the nodeName of the pod from Pod object(v1.Pod), but for first task in our case will not have nodeName in pod(v1.Pod) object, since it is not yet sent for binding, it will be in pending state, so for second task, we will get error that no node with nill value(empty string)

`Calculate Inter Pod Affinity Priority Failed because of Error: failed to find node <>`

So both not being able to switch from Pending to Running state since the `occupied value will be less than MinAvailable`, so our gang scheduling will not work with K8s InterPod Affinity Priority logic.
In K8s Pod Affinity Predicates, they use `topologyPairsAntiAffinityPodsMap` in their `predicateMetaData`, and run predicates for the task, so predicates was working fine with gang scheduling, but InterPodAffinity Priority function in K8s will not support gang scheduling model.

k82cn: If so, we need to find a solution. Affiinity to itself is a valid case.

k82cn: Maybe we can deepcopy pods and set its nodename by TaskInfo when listing. BTW, only deep copy the pod that NodeName is empty but TaskInfo.NodeName is NOT empty.

thandayuthapani: But that listing is done by InterPod Affinity code provided by K8s, if we have to change the code, we will have to bring it out of vendor and place it in kube-batch.  We just send current pod object(v1.Pod), list of nodes(both schedulercache.NodeInfo(as map) and v1.Nodes(as slice)).

k82cn: > But that listing is done by InterPod Affinity code provided by K8s

It will use lister from our code, right?

```
mapFn := priorities.NewInterPodAffinityPriority(cn, nl, pl, v1.DefaultHardPodAffinitySymmetricWeight)
```

thandayuthapani: Yes, have made change in our lister code, and Affinity to itself case also passes now.  And have changed predicate Pod affinity test case to the one that was existing before.

pkg/scheduler/plugins/prioritize/prioritize.go
```diff
@@ -0,0 +1,206 @@
...
+	var pods []*v1.Pod
+	for _, job := range pl.session.Jobs {
+		for status, tasks := range job.TaskStatusIndex {
+			if !api.AllocatedStatus(status) {
+				continue
+			}
+
+			for _, task := range tasks {
+				if selector.Matches(labels.Set(task.Pod.Labels)) {
+					pod := task.Pod.DeepCopy()
```
k82cn: we only need DeepCopy the pod that `task.NodeName != Pod.NodeName`.

pkg/scheduler/plugins/prioritize/prioritize.go
```diff
@@ -0,0 +1,206 @@
...
+	var pods []*v1.Pod
+	for _, job := range pl.session.Jobs {
+		for status, tasks := range job.TaskStatusIndex {
+			if !api.AllocatedStatus(status) {
+				continue
+			}
+
+			for _, task := range tasks {
+				if podFilter(task.Pod) && selector.Matches(labels.Set(task.Pod.Labels)) {
+					pod := task.Pod.DeepCopy()
```
k82cn: ditto

test/e2e/predicates.go
k82cn: LGTM overall, please only do DeepCopy when `task.NodeName != pod.NodeName`.

pkg/scheduler/plugins/prioritize/prioritize.go
thandayuthapani: In interPodAffinity Priority code, we do not use PodLister's `List` and `FilteredList` function, list is done implicitly in the interPodAffinity code, where is ranges over every node and get pods from nodeinfo object, and calls processPod function, which uses only `GetNodeInfo` function alone.

```
processNode := func(i int) {
		nodeInfo := nodeNameToInfo[allNodeNames[i]]
                ......
		......		
               for _, existingPod := range nodeInfo.Pods() {
			if err := processPod(existingPod); err != nil {
```
  
In interPodAffinity Priority code of K8s, uses 
`existingPodNode, err := ipa.info.GetNodeInfo(existingPod.Spec.NodeName)`
function, since `existingPod.Spec.NodeName` value was nil, we were getting error, so have made changes only in `func (c *cachedNodeInfo) GetNodeInfo(name string) (*v1.Node, error)` function alone, other listers are same as there in Predicate plugin.  Changes has been done only to `GetNodeInfo` function for fixing that case.  Not in `FilteredList` and `List`, those are same as that of present in predicate plugin.  `FilteredList` and `List` function is not used in interPodAffinity priority code, it is defined only because that struct should implement a interface.  Changes in those function will not affect the behaviour of priority function.

test/e2e/predicates.go
thandayuthapani: We don't use `List` and `FilteredList` function in interPodAffinity priority code.  Even if that check is included, it will not affect behaviour of priority, so have just implemented the same as that of implemented in predicate plugin.

pkg/scheduler/plugins/prioritize/prioritize.go
k82cn: if `task.NodeName == pod.NodeName`, append pod as before;  if `task.NodeName != pod.NodeName`, do DeepCopy and then append copied pod. Do you mean that does not work?

thandayuthapani: That `List` function where you are asking me to change, is not used by interPodAffinity priority code.

k82cn: what's different? If NodeName not change, why do we do DeepCopy? priority and predicate only read pod info.

thandayuthapani: Have changed the implementation, according to your comments

k82cn:
/lgtm

Thanks for your contribution :)

PR #626, opened 2019-03-08
Fix helm deployment

thandayuthapani:
Kube-batch crashes when deployed using helm.  `--enable-namespace-as-queue` has been removed, but helm deployment uses it. So have removed that flag from deployment.


k82cn:
@thandayuthapani we need to cherry pick this to release-0.4

PR #631, opened 2019-03-11
Add helm hook crd-install to fix helm install error

hex108:
There is an error when running `helm install $GOPATH/src/github.com/kubernetes-sigs/kube-batch/deployment/kube-batch --namespace kube-system`.

Error message:

> $ helm install deployment/kube-batch --namespace kube-system
Error: validation failed: unable to recognize "": no matches for kind "Queue" in version "scheduling.incubator.k8s.io/v1alpha1"

It is because helm does not guarantee that `deployment/kube-batch/templates/scheduling_v1alpha1_queue.yaml` will be executed before `deployment/kube-batch/templates/default.yaml`. However we could add a helm hook [`crd-install`](https://github.com/helm/helm/blob/master/docs/charts_hooks.md) to guarantee it.

Fixes #

**Release note**:

```release-note
Add helm hook crd-install to fix helm install error
```

### DIFF
--- a/deployment/kube-batch/templates/scheduling_v1alpha1_queue.yaml
+++ b/deployment/kube-batch/templates/scheduling_v1alpha1_queue.yaml
@@ -2,6 +2,8 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: queues.scheduling.incubator.k8s.io
+  annotations:
+    "helm.sh/hook": "crd-install"
 spec:
   group: scheduling.incubator.k8s.io
   names:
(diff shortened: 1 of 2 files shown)


deployment/kube-batch/templates/scheduling_v1alpha1_queue.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: queues.scheduling.incubator.k8s.io
+  annotations:
+    "helm.sh/hook": "crd-install"
```
k82cn, line 6: can you also help to add this hook for `PodGroup`?

hex108: The YAML files in the chart do not use `PodGroup`, so it seems unnecessary? I'll add it if needed :)

hex108: Added it in case that some logic(e.g. logic in the `kube-batch` deployment) depends on it.

k82cn: thanks :) Just in case we add some other dependency on that.

PR #638, opened 2019-03-13
Take init containers into account when getting pod resource request

hex108:
Take init containers into account when getting pod resource request. The code refers `k8s.io/kubernetes/pkg/scheduler/algorithm/predicates/predicates.go#GetResourceRequest`.

Fixes #

**Release note**:

```release-note
Take init containers into account when getting pod resource request
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -143,7 +143,7 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 			selectedNodes := util.SelectBestNode(nodeScores)
 			for _, node := range selectedNodes {
 				// Allocate idle resource to the task.
-				if task.Resreq.LessEqual(node.Idle) {
+				if task.InitResreq.LessEqual(node.Idle) {
 					glog.V(3).Infof("Binding Task <%v/%v> to node <%v>",
 						task.Namespace, task.Name, node.Name)
 					if err := ssn.Allocate(task, node.Name); err != nil {
@@ -162,9 +162,9 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 				}
 
 				// Allocate releasing resource to the task if any.
-				if task.Resreq.LessEqual(node.Releasing) {
+				if task.InitResreq.LessEqual(node.Releasing) {
 					glog.V(3).Infof("Pipelining Task <%v/%v> to node <%v> for <%v> on <%v>",
-						task.Namespace, task.Name, node.Name, task.Resreq, node.Releasing)
+						task.Namespace, task.Name, node.Name, task.InitResreq, node.Releasing)
 					if err := ssn.Pipeline(task, node.Name); err != nil {
 						glog.Errorf("Failed to pipeline Task %v on %v in Session %v",
 							task.UID, node.Name, ssn.UID)
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -38,7 +38,10 @@ type TaskInfo struct {
 	Name      string
 	Namespace string
 
+	// Resreq is the resource that used when task running.
 	Resreq *Resource
+	// InitResreq is the resource that used to launch a task.
+	InitResreq *Resource
 
 	NodeName    string
 	Status      TaskStatus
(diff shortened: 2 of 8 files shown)


k82cn:
/approve


LGTM overall, let's get it merged when CI is happy :)

hex108:
@k82cn I'll look into it.

I checked the code of k8s scheduler again, and found it only consider init containers' resource request for the predicate `PodFitsResource`, it will not calculate them when adding it to node's total requested resource. Need we be consistent with this behavior?

Another question: is there any reason to use float64 instead of int64 to record pod resource?

k82cn:
> Need we be consistent with this behavior?

Yes; we allocate resources in `allocate` action so `PodFitsResource` is not necessary; but initContainer is still necessary :)

> Another question: is there any reason to use float64 instead of int64 to record pod resource?

We used to use CPU instead of MillCPU, so float64 is there.

hex108:
> > Need we be consistent with this behavior?
> 
> Yes; we allocate resources in `allocate` action so `PodFitsResource` is not necessary; but initContainer is still necessary :)

Addressed. PTAL.

> 
> > Another question: is there any reason to use float64 instead of int64 to record pod resource?
> 
> We used to use CPU instead of MillCPU, so float64 is there.

Get it. Thanks.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -141,9 +141,10 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 				}
 			}
 			selectedNodes := util.SelectBestNode(nodeScores)
+			req := api.GetPodResourceRequest(task.Pod)
```
k82cn: `task.Resreq` already includes `initContainer`, why do we need to re-calculate it again?

hex108: `task.Resreq` is calculated by `GetPodResourceWithoutInitContainers`, it does not contain `initContainer`. We only need consider `initContainer` when `Allocate`, do I understand it correctly?

k82cn: We also need Resreq in reclaim, preemption and backfill. And predicates&priorities will use Pods instead of task info :)

hex108: Addressed. Thanks!

k82cn: hm.... can we update `task.Resreq` directly, so we do not need to update different actions :)

hex108: I updated it at first time. However a lot of functions(e.g. `jobinfo.AddTaskInfo`, `NodeInfo.AddTaskInfo`) will use `task.Resq` that does not contain `init container`. Maybe we could add another variable `ResqWithInitContainers` in `TaskInfo`?

k82cn: ~~If so, prefer to update related code to consider `initContainer` :)~~

hm.... maybe we need to distinguish them. In nodeInfo, it's not necessary to consider `initContainer` if it's in running state.

pkg/scheduler/api/job_info.go
```diff
@@ -38,7 +38,12 @@ type TaskInfo struct {
 	Name      string
 	Namespace string
 
-	Resreq *Resource
+	// Resreq is pod resource request, and it does not contain init containers resource request.
+	// ResreqWithInitContainers is also pod resource request, however it contains init containers resource request.
+	// To be consistent with kubernetes default scheduler, ResreqWithInitContainers is only used for predicates
+	// of actions (e.g.allocate, backfill, preempt, reclaim), please use Resreq for other cases.
+	Resreq                   *Resource
+	ResreqWithInitContainers *Resource
```
k82cn: How about `InitResreq`? The `Resreq` is the resource that used when task running; and the `InitResreq` is the resource that used to launch a task.

hex108: OK, it is shorter :) Addressed.

pkg/scheduler/api/job_info.go
```diff
@@ -61,25 +64,22 @@ func getJobID(pod *v1.Pod) JobID {
...
+		UID:        TaskID(pod.UID),
+		Job:        jobID,
+		Name:       pod.Name,
+		Namespace:  pod.Namespace,
+		NodeName:   pod.Spec.NodeName,
+		Status:     getTaskStatus(pod),
+		Priority:   1,
+		Pod:        pod,
+		Resreq:     req,
+		InitResreq: initResreq,
```
k82cn, line 82: we also need to update `Clone()` :)

hex108: Sorry for it, forget to run test. Updated.

hex108:
Test errors are because namespace test is not deleted.

PR #642, opened 2019-03-13
Add event when task is scheduled

Rajadeepan:
This PR add a event when task is scheduled


pkg/scheduler/cache/cache.go
```diff
@@ -424,6 +424,7 @@ func (sc *SchedulerCache) Bind(taskInfo *kbapi.TaskInfo, hostname string) error
 		if err := sc.Binder.Bind(p, hostname); err != nil {
 			sc.resyncTask(task)
 		}
+		sc.Recorder.Eventf(p, v1.EventTypeNormal, "Scheduled", "Successfully assigned %v/%v to %v", p.Namespace, p.Name, hostname)
```
k82cn: If err != nil, we should not record event.

PR #646, opened 2019-03-18
Allocate: when failing to allocate for one task, continue to allocate…

hex108:
… for next tasks

There might be many different tasks with different resource requests in a same job, so it does not mean there are no suitable nodes that meet next tasks's requirements when it fails to allocate for one task.

However I have some concerns for this PR:
* It might reduce performance because it will continue to allocate even when next tasks' resource requirement are same with the failed allocated task. Maybe eCache is a way for performance improvement?

There is also a common problem(not sure it is a valid problem :)) I am not sure whether there is a good solution for it: suppose there is a job with 3 tasks for kind A, 4 task of kind B, minMember is 3 which requires 1 task of kind A, and 2 tasks of kind B. For current algorithm, it might allocate 3 tasks of kind A, and find job is ready, then dispatch the job. However the job is not ready actually.

Fixes #

**Release note**:

```release-note
Allocate: when failing to allocate for one task, continue to allocate for next tasks.
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -107,7 +107,6 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 			nodeScores := map[int][]*api.NodeInfo{}
 
 			task := tasks.Pop().(*api.TaskInfo)
-			assigned := false
 
 			glog.V(3).Infof("There are <%d> nodes for Job <%v/%v>",
 				len(ssn.Nodes), job.Namespace, job.Name)
@@ -151,7 +150,6 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 							task.UID, node.Name, ssn.UID, err)
 						continue
 					}
-					assigned = true
 					break
 				} else {
 					//store information about missing resources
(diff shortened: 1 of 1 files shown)


k82cn:
Honestly, we used to have a this before, but I revert it because task-priority feature dependent on that, e.g. Spark.

If we want to support that, we need a parameter or option to `PodGroup` on whether continue or not.

hex108:
> Honestly, we used to have a this before, but I revert it because task-priority feature dependent on that, e.g. Spark.
> 
> If we want to support that, we need a parameter or option to `PodGroup` on whether continue or not.

Get it. Yes, it will be better to add a parameter or option. 

What's your option for the following problem? It could not be solved by task-priority.
> There is also a common problem(not sure it is a valid problem :)) I am not sure whether there is a good solution for it: suppose there is a job with 3 tasks for kind A, 4 task of kind B, minMember is 3 which requires 1 task of kind A, and 2 tasks of kind B. For current algorithm, it might allocate 3 tasks of kind A, and find job is ready, then dispatch the job. However the job is not ready actually.

k82cn:
> What's your option for the following problem? It could not be solved by task-priority.

hm... I'm thinking to have minNumber for different tasks, e.g. a map of taskSpec name to minNumber. But we did not get such kind of requirement from user right now.

hex108:
> > What's your option for the following problem? It could not be solved by task-priority.
> 
> hm... I'm thinking to have minNumber for different tasks, e.g. a map of taskSpec name to minNumber. But we did not get such kind of requirement from user right now.

We could start to work on it when there is requirement.

k82cn:
I'm going to close this PR as we discussed. If any more case, let's open a new one :)

PR #655, opened 2019-03-21
Do graceful eviction using default policy

hex108:
According to `k8s.io/kuberentes/pkg/registry/core/pod/strategy.go#CheckGracefulDelete`, if `DeleteOptions` is nil, it will check `pod`'s filed `pod.Spec.TerminationGracePeriodSeconds`. Maybe we could just leave it as nil?

Fixes #

**Release note**:

```release-note
NONE
```


k82cn:
I'm ok with that; if anyone has requirement on different setting, we can open another PR for that.

/lgtm
/approve

Issue #665, opened 2019-03-27
Wrong calculation for queue deserved in proportion plugin


zionwu:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
The kube-batch crash with following error message:
```
panic: Resource is not sufficient to do operation: <cpu 408833.33, memory 5346978668544.00, GPU 0.00> sub <cpu 608000.00, memory 5153960755200.00, GPU 0.00> [recovered]
	panic: Resource is not sufficient to do operation: <cpu 408833.33, memory 5346978668544.00, GPU 0.00> sub <cpu 608000.00, memory 5153960755200.00, GPU 0.00>

goroutine 246 [running]:
github.com/kubernetes-sigs/kube-batch/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/Users/klaus/Workspace/kb_ws/src/github.com/kubernetes-sigs/kube-batch/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:58 +0x108
panic(0x111cb00, 0xc00102a870)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api.(*Resource).Sub(0xc001003760, 0xc0010037e0, 0xc000564458)
	/Users/klaus/Workspace/kb_ws/src/github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api/resource_info.go:108 +0x10e
github.com/kubernetes-sigs/kube-batch/pkg/scheduler/plugins/proportion.(*proportionPlugin).OnSessionOpen(0xc000f78360, 0xc00039a9c0)
	/Users/klaus/Workspace/kb_ws/src/github.com/kubernetes-sigs/kube-batch/pkg/scheduler/plugins/proportion/proportion.go:138 +0xefc
github.com/kubernetes-sigs/kube-batch/pkg/scheduler/framework.OpenSession(0x143e180, 0xc0002a01c0, 0xc000e4e280, 0x5, 0x8, 0x0)
	/Users/klaus/Workspace/kb_ws/src/github.com/kubernetes-sigs/kube-batch/pkg/scheduler/framework/framework.go:36 +0x275
github.com/kubernetes-sigs/kube-batch/pkg/scheduler.(*Scheduler).runOnce(0xc000395ce0)
```

[...]

PR #666, opened 2019-03-27
Fix wrong caculation for deserved in proportion plugin

zionwu:
Fix issue https://github.com/kubernetes-sigs/kube-batch/issues/665

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -124,6 +124,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 				continue
 			}
 
+			oldDeserved := attr.deserved.Clone()
 			attr.deserved.Add(remaining.Clone().Multi(float64(attr.weight) / float64(totalWeight)))
 			if !attr.deserved.LessEqual(attr.request) {
 				attr.deserved = helpers.Min(attr.deserved, attr.request)
@@ -134,7 +135,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			glog.V(4).Infof("The attributes of queue <%s> in proportion: deserved <%v>, allocate <%v>, request <%v>, share <%0.2f>",
 				attr.name, attr.deserved, attr.allocated, attr.request, attr.share)
 
-			deserved.Add(attr.deserved)
+			deserved.Add(attr.deserved.Clone().Sub(oldDeserved))
 		}
 
 		remaining.Sub(deserved)


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -134,7 +135,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			glog.V(4).Infof("The attributes of queue <%s> in proportion: deserved <%v>, allocate <%v>, request <%v>, share <%0.2f>",
 				attr.name, attr.deserved, attr.allocated, attr.request, attr.share)
 
-			deserved.Add(attr.deserved)
+			deserved.Add(queueDeserved)
```
hex108: If `!attr.deserved.LessEqual(attr.request)`, `queueDeserved` need be changed to `attr.deserved` increased value.

hex108:
@zionwu Thanks for the patch! Could you please help add a test for it?

pkg/scheduler/plugins/proportion/proportion.go
zionwu: Sorry, I did not get it. Could you explain a bit more?

I think the `queueDeserved` is already the increased value. In the for loop we add it to `attr.deserved`  and `deserved`. After the for loop, we calculate the `remaining` by subtracting  `deserved`

hex108: Suppose that `queueDeserved` is `{cpu: 2, mem: 2G}`, `attr.deserved` is `{cpu: 1, mem: 1G}` and `attr.request` is `{cpu: 2, mem: 2G}`, `attr.deserved` will be `{cpu: 3, mem: 3G}` after `attr.deserved.Add(queueDeserved)`(line 128). Then `attr.deserved` will be `{cpu: 2, mem: 2G}` after `attr.deserved = helpers.Min(attr.deserved, attr.request)`(line 130). So `attr.deserved`'s increased value is actually `{cpu: 1, mem: 1G}`, and we just need add `{cpu: 1, mem: 1G}` instead of `{cpu: 2, mem: 2G}` to `deserved`.

zionwu: got it, thanks.  Let me update it.

zionwu: Updated.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -124,17 +124,20 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 				continue
 			}
```
hex108, line 126: Add: oldDeserved = attr.deserved.Clone()

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -124,17 +124,20 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
...
 				attr.deserved = helpers.Min(attr.deserved, attr.request)
 				meet[attr.queueID] = struct{}{}
+				increasedDeserved = attr.deserved.Clone().Sub(oldDeserved)
 			}
 			pp.updateShare(attr)
 
 			glog.V(4).Infof("The attributes of queue <%s> in proportion: deserved <%v>, allocate <%v>, request <%v>, share <%0.2f>",
 				attr.name, attr.deserved, attr.allocated, attr.request, attr.share)
 
-			deserved.Add(attr.deserved)
```
hex108, line 137: Changed to: deserved.Add(attr.deserved.Clone().sub(oldDeserved))

pkg/scheduler/plugins/proportion/proportion.go
hex108: Thanks! You code works. Just give some suggestion.

pkg/scheduler/plugins/proportion/proportion.go
zionwu: Updated

pkg/scheduler/plugins/proportion/proportion.go
zionwu: Updated

hex108:
lgtm except missing tests...

/assign @k82cn

PR #705, opened 2019-04-04
Add type Argument with some common parse function

### LINKED ISSUE #704: Add type Argument with some common parse function

> Uncomment only one, leave it on its own line:
>
> /kind bug

 /kind feature

**What happened**:
PluginBuilder has function definition with code  Introduced in  #637 
> type PluginBuilder func(map[string]string) Plugin

Every plugin use it's own method to parse the arguments.

**What you expected to happen**:
a common support for arguments parse

**How to reproduce it (as minimally and precisely as possible)**:
I wanna to use a special type to replace it and add some common argument parse function.
> type Arguments map[string]string
> type PluginBuilder func(Arguments) Plugin

add function
> func (a Arguments) GetInt(ptr *int, key string)

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): N/A
- Cloud provider or hardware configuration:N/A
- OS (e.g. from /etc/os-release):N/A
- Kernel (e.g. `uname -a`):N/A
- Install tools:N/A
- Others:N/A

lmzqwer2:
1. add special type for plugin arguments
2. add common parse function GetInt

Fixes #704

**Release note**:

```release-note

```

### DIFF
--- a/pkg/scheduler/framework/arguments.go
+++ b/pkg/scheduler/framework/arguments.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"strconv"
+
+	"github.com/golang/glog"
+)
+
+type Arguments map[string]string
+
+func (a Arguments) GetInt(ptr *int, key string) {
+	if ptr == nil {
+		return
+	}
+
+	argv, ok := a[key]
+	if !ok || argv == "" {
+		return
+	}
+
+	value, err := strconv.Atoi(argv)
+	if err != nil {
+		glog.Warningf("Could not parse argument: %s for key %s, with err %v", argv, key, err)
+		return
+	}
+
+	*ptr = value
+}
--- a/pkg/scheduler/framework/arguments_test.go
+++ b/pkg/scheduler/framework/arguments_test.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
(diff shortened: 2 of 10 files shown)


pkg/scheduler/framework/arguments.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"strconv"
+
+	"github.com/golang/glog"
+)
+
+type Arguments map[string]string
+
+func (a Arguments) GetInt(ptr *int, key string) {
```
k82cn, line 27: return value instead of pointer.

pkg/scheduler/framework/arguments.go
```diff
@@ -0,0 +1,28 @@
...
+
+	argv, ok := a[key]
+	if !ok || argv == "" {
+		return
+	}
+
+	value, err := strconv.Atoi(argv)
+	if err != nil {
+		glog.Warningf("Could not parse argument: %s for key %s, with err %v", argv, key, err)
+		return
```
k82cn, line 40: ~~the return value is `(int, error)`~~

It's ok to me :)

pkg/scheduler/framework/arguments.go
```diff
@@ -0,0 +1,28 @@
+package framework
```
k82cn, line 17: add license there.

pkg/scheduler/framework/arguments.go
lmzqwer2: so, I think the definition should be.
```
var ERRNOTFOUND=errors.New("ARGUMENT NOT FOUND")

func (a Arguments) GetInt(key string) (value int, err error) { ... }
```
when the value of this key is not found, the return value should be 0, ERRNOTFOUND
when the value is in invalid format of int, the return value should be 0, err of parsing

pkg/scheduler/framework/arguments.go
lmzqwer2: copy license from other file.

And I think the license may be outdated now? 
```
Copyright 2018 The Kubernetes Authors
```

pkg/scheduler/framework/arguments.go
lmzqwer2: ok. :)

pkg/scheduler/framework/arguments.go
hex108: Could use 2019 for new added files

lmzqwer2: New added files (arguments.go arguments_test.go) is now in license of 2019.

PR #707, opened 2019-04-04
Add resource predicates for tasks

### LINKED ISSUE #694: Add resource predicates for tasks

/kind feature

**Description**:

Currently, we did not use resource predicates from upstream because of releasing resources; but it makes unnecessary sort and check on the full nodes. It's better to have a predicates to check `Task.Resreq` against `Node.Idle` and `Node.Release` resources.

thandayuthapani:
Add resource predicates for tasks
**Which issue(s) this PR fixes**:
Fixes https://github.com/kubernetes-sigs/kube-batch/issues/694

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -144,9 +144,16 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 					glog.V(3).Infof("Predicates failed for task <%s/%s> on node <%s>: %v",
 						task.Namespace, task.Name, node.Name, err)
 				} else {
-					workerLock.Lock()
-					predicateNodes = append(predicateNodes, node)
-					workerLock.Unlock()
+					clonedNode := node.Idle.Clone()
+					// Check for Resource Predicate
+					if task.InitResreq.LessEqual(clonedNode.Add(node.Releasing)) {
+						workerLock.Lock()
+						predicateNodes = append(predicateNodes, node)
+						workerLock.Unlock()
+					} else {
+						glog.V(3).Infof("task <%s/%s> ResourceFit failed on node <%s>",
+							task.Namespace, task.Name, node.Name)
+					}
 				}
 			}
 			workqueue.ParallelizeUntil(context.TODO(), 16, len(allNodes), checkNode)


pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -199,6 +199,13 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				node.Name, task.Namespace, task.Name)
 		}
 
+		// Predicate checks whether enough resource is available for the task in that node
+		clonedNode := node.Clone()
+		if !task.Resreq.LessEqual(clonedNode.Idle.Add(node.Releasing)) {
```
hex108: Do we need use `task.InitResreq` here?

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -199,6 +199,13 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			return fmt.Errorf("task <%s/%s> ResourceFit failed on node <%s>",
+				node.Name, task.Namespace, task.Name)
```
hex108: `node.Name` should be put at last variable

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -199,6 +199,13 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				node.Name, task.Namespace, task.Name)
 		}
 
+		// Predicate checks whether enough resource is available for the task in that node
+		clonedNode := node.Clone()
```
hex108: Just need clone `Idle` resource?

pkg/scheduler/plugins/predicates/predicates.go
k82cn: we also need to check `Releasing` resources.

thandayuthapani: done

pkg/scheduler/plugins/predicates/predicates.go
thandayuthapani: done

pkg/scheduler/plugins/predicates/predicates.go
thandayuthapani: done

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -199,6 +199,13 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				node.Name, task.Namespace, task.Name)
 		}
 
+		// Predicate checks whether enough resource is available for the task in that node
+		clonedNode := node.Idle.Clone()
+		if !task.InitResreq.LessEqual(clonedNode.Add(node.Releasing)) {
+			return fmt.Errorf("task <%s/%s> ResourceFit failed on node <%s>",
+				node.Name, task.Namespace, node.Name)
```
k82cn: should be `task.Namespace, task.Name, node.Name`.

thandayuthapani:
If we add Resource predicate check in predicate plugin, it affects preemption also, so have moved that check to Allocate alone

pkg/scheduler/plugins/predicates/predicates.go
thandayuthapani: done

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -144,9 +144,16 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 					glog.V(3).Infof("Predicates failed for task <%s/%s> on node <%s>: %v",
 						task.Namespace, task.Name, node.Name, err)
 				} else {
-					workerLock.Lock()
-					predicateNodes = append(predicateNodes, node)
-					workerLock.Unlock()
+					clonedNode := node.Idle.Clone()
+					// Check for Resource Predicate
+					if task.InitResreq.LessEqual(clonedNode.Add(node.Releasing)) {
```
k82cn, line 149: let's check this before other predicates; if predicates failed, we can return it directly.

k82cn:
@thandayuthapani , thanks very much for your effort. I resolved conflict at #717 to catch up internal schedule, sorry for that.

Issue #699, opened 2019-04-02
Replace NodeOrder with BestNode


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

Currently, we're using similar way with upstream to sort nodes for a task to get the best node; but the sort action is not necessary as we just want to get the best node after predicates. It's better to get best node directly instead of sort, the time complexity will be reduced from `O(nlogn)` to `O(n)`.

hex108:
If the best node doest not fits, we need pick next one, so will it be better to return a `PriorityQueue`?

k82cn:
why best node does not fit after predicates?

hex108:
@k82cn  If I understand correctly, the predicates does not contain resource predicates, so there is some resource compare logic in https://github.com/kubernetes-sigs/kube-batch/blob/master/pkg/scheduler/actions/allocate/allocate.go#L168. Need we enhance the prediates?

Another situation: administrator perhaps does not enable `predicate` plugin.

k82cn:
> Need we enhance the prediates?

yes, refer to https://github.com/kubernetes-sigs/kube-batch/issues/694 .

> Another situation: administrator perhaps does not enable `predicate` plugin.

Yes, but scheduler can not do anything to enhance that even we have a ordered list :)

hex108:
> > Need we enhance the prediates?
> 
> yes, refer to #694 .

Thanks for it!

> > Another situation: administrator perhaps does not enable `predicate` plugin.
> 
> Yes, but scheduler can not do anything to enhance that even we have a ordered list :)

There is resource predicate in https://github.com/kubernetes-sigs/kube-batch/blob/master/pkg/scheduler/actions/allocate/allocate.go#L168. If we fix #694, we could delete those code. However after we add resource predicate in predicates, if there are no nodes satisfying the task, we will try to allocate for next job and we could not pipeline task(it works like reserving resource), big tasks might starve.

k82cn:
> If we fix #694, we could delete those code. 

In resource predicates, we need to check both idel & releasing to avoid duplicated preemption; the logic is necessary here as the action maybe different when using resources.

k82cn:
btw, we only need this in `allocate` action.

PR #719, opened 2019-04-08
NodeOrderFn return float value

### LINKED ISSUE #708: Change return value of NodeOrderFn from int to float

> Uncomment only one, leave it on its own line:
>
> /kind bug

 /kind feature

**What happened**:
definition of NodeOrderFn is
```
type NodeOrderFn func(*TaskInfo, *NodeInfo) (int, error)
```

but the **int** type would lost precision when two node is similar.

**What you expected to happen**:
we can distinguish two node with similar priority

**How to reproduce it (as minimally and precisely as possible)**:
I wanna to change the type of return value from **int** to **float** 
```
type NodeOrderFn func(*TaskInfo, *NodeInfo) (float64, error)
```

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): N/A
- Cloud provider or hardware configuration:N/A
- OS (e.g. from /etc/os-release):N/A
- Kernel (e.g. `uname -a`):N/A
- Install tools:N/A
- Others:N/A

lmzqwer2:
As I said in issue #708, changing the return value of NodeOrderFn from float to int would be better.

The ability of NodeOrderFn remains unchanged now, because functions like leastReq still returns int type and int type could change to float type safely.

But the code after this would benefit from it.
 
Fixes #708

1. Just change the return value

**Release note**:

```release-note

```

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -220,7 +220,7 @@ func calculateWeight(args framework.Arguments) priorityWeight {
 }
 
 func (pp *nodeOrderPlugin) OnSessionOpen(ssn *framework.Session) {
-	nodeOrderFn := func(task *api.TaskInfo, node *api.NodeInfo) (int, error) {
+	nodeOrderFn := func(task *api.TaskInfo, node *api.NodeInfo) (float64, error) {
 
 		weight := calculateWeight(pp.pluginArguments)
 
@@ -244,7 +244,7 @@ func (pp *nodeOrderPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		nodeInfo := cache.NewNodeInfo(node.Pods()...)
 		nodeInfo.SetNode(node.Node)
-		var score = 0
+		var score = 0.0
 
 		//TODO: Add ImageLocalityPriority Function once priorityMetadata is published
 		//Issue: #74132 in kubernetes ( https://github.com/kubernetes/kubernetes/issues/74132 )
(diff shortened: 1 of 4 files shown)


pkg/scheduler/util/scheduler_helper.go
```diff
@@ -75,13 +75,13 @@ func PrioritizeNodes(task *api.TaskInfo, nodes []*api.NodeInfo, fn api.NodeOrder
 }
 
 // SelectBestNode returns nodes by order of score
-func SelectBestNode(nodeScores map[int][]*api.NodeInfo) []*api.NodeInfo {
+func SelectBestNode(nodeScores map[float64][]*api.NodeInfo) []*api.NodeInfo {
```
k82cn, line 78: hm... it's better to work with https://github.com/kubernetes-sigs/kube-batch/issues/699 :)

/cc @hex108

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -255,23 +255,23 @@ func (pp *nodeOrderPlugin) OnSessionOpen(ssn *framework.Session) {
 			return 0, err
 		}
 		// If leastReqWeight in provided, host.Score is multiplied with weight, if not, host.Score is added to total score.
-		score = score + (host.Score * weight.leastReqWeight)
+		score = score + float64(host.Score*weight.leastReqWeight)
```
hex108, line 258: It gets `int` at each step, where does it get `float` value?

lmzqwer2: We don't have a function with float value now.

But I think the calculation of leastReq should return float and others may meet the same situation. So I just provide a way to choose the best fist node but not a set of similar node with the same score.

pkg/scheduler/util/scheduler_helper.go
hex108: ok, I'll send PR after this.

Issue #591, opened 2019-02-14
Increase code coverage by e2e


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

Currently, there're only ~10 e2e test for kube-batch which are not enough for the quality.

PR #726, opened 2019-04-09
Adding e2e test case for resource limit checks

shivramsrivastava:
Added a new test case to check the resource limit calculation by the scheduler.
This test creates multiple podgroups and exhausts the cluster resource and then tries to create other podgroups which will stay unscheduled.

#591 

**Release note**:

```release-note

```

### DIFF
--- a/test/e2e/predicates.go
+++ b/test/e2e/predicates.go
@@ -18,10 +18,13 @@ package e2e
 
 import (
 	"fmt"
-	"github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
+	kubev1 "github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeletapi "k8s.io/kubernetes/pkg/kubelet/apis"
@@ -259,7 +262,7 @@ var _ = Describe("Predicates E2E Test", func() {
 		// These jobs will exhaust 70% of the CPU on that node.
 		// Using node-affinity assign the job per node.
 
-		cpuPodGrpList := []*v1alpha1.PodGroup{}
+		cpuPodGrpList := []*kubev1.PodGroup{}
 		for nodeName, cpu := range nodeToAllocatableMapCpu {
 			requestedCPU := cpu * 7 / 10
 			cpuJob := &jobSpec{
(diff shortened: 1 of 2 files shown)


test/e2e/predicates.go
```diff
@@ -190,4 +193,210 @@ var _ = Describe("Predicates E2E Test", func() {
...
+			}
+		}
+
+		pods, err := context.kubeclient.CoreV1().Pods(metav1.NamespaceAll).List(metav1.ListOptions{})
+		checkError(context, err)
+		for _, pod := range pods.Items {
+			_, found := nodeToAllocatableMapCpu[pod.Spec.NodeName]
+			_, foundMem := nodeToAllocatableMemMap[pod.Spec.NodeName]
+			if found && foundMem && pod.Status.Phase != v1.PodSucceeded && pod.Status.Phase != v1.PodFailed {
+				nodeToAllocatableMapCpu[pod.Spec.NodeName] -= getRequestedCPU(pod)
```
hex108, line 255: Maybe we could use `GetPodResourceRequest(pod)` to get pod resource request?

test/e2e/predicates.go
```diff
@@ -17,11 +17,14 @@ limitations under the License.
 package e2e
 
 import (
+	"fmt"
+	"github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
```
hex108: It is better to be put at a separate part.

test/e2e/util.go
```diff
@@ -46,7 +46,7 @@ import (
 	kbapi "github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api"
 )
 
-var oneMinute = 1 * time.Minute
+var oneMinute = 5 * time.Minute
```
hex108: If we need more time, we also need change the name `oneMinute`.

test/e2e/predicates.go
shivramsrivastava: ```GetPodResourceRequest``` method return the max of the pod resource.
We need the sum of the resource in this test case.

test/e2e/predicates.go
shivramsrivastava: i didn't get this comment.
You mean we need a better alias name for this package?
i have added a new alias for this.

test/e2e/util.go
shivramsrivastava: Addressed this review comment.
Added a new variable and also added a two new helper methods with timeout.

test/e2e/predicates.go
hex108: sorry, it is better to be:

```
import (
	"fmt"
     
         . "github.com/onsi/ginkgo"
         . "github.com/onsi/gomega"

         "k8s.io/api/core/v1"
+        "k8s.io/apimachinery/pkg/api/resource"
          metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+        kubeletapi "k8s.io/kubernetes/pkg/kubelet/apis"
          schedulerapi "k8s.io/kubernetes/pkg/scheduler/api"

+	 "github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
)
```

shivramsrivastava: Done.
Thanks for the review.

PR #730, opened 2019-04-10
Fix wrong calculation for queue deserved in proportion plugin

### LINKED ISSUE #729: Deserved attr is not correctly calculated in proportion plugin
 /kind bug

**What happened**:

I have a job  not getting scheduled while there is enough resources in the cluster. I looked into the log and find that  in allocate action, the queue is marked as overused:

> I0410 04:06:16.001089       1 allocate.go:72] Queue <11073333> is overused, ignore it.
I0410 04:06:16.001094       1 allocate.go:72] Queue <11073333> is overused, ignore it

The reason why the queue is overused is that in proportion plugin, the deserved value for the queue is not correctly calculated:
```golang		
			attr.deserved.Add(remaining.Clone().Multi(float64(attr.weight) / float64(totalWeight)))
			if !attr.deserved.LessEqual(attr.request) {
				attr.deserved = helpers.Min(attr.deserved, attr.request)
				meet[attr.queueID] = struct{}{}
			}
```
For example, the attr.deserved is `<cpu 523750.00, memory 3076011404288.00, GPU 3750.00>` and attr.request is `<cpu 608000.00, memory 5153960755200.00, GPU 0.00>`, 
` !attr.deserved.LessEqual(attr.request)` return true and the queue is set to meet and is not allocated enough resources.

I think we should use ` attr.request.LessEqual(attr.deserved)`  instead. 

Another problem is that the calculation of the total increased deserved:
```
			deserved.Add(attr.deserved.Clone().Sub(oldDeserved))

```
We assume that the attr.deserved is greater than oldDeserved, which is wrong，for example,  `oldDeserved` can be `<cpu 523750.00, memory 3076011404288.00, GPU 3750.00>` and `attr.deserved` can be `<cpu 500750.00, memory 5076011404288.00, GPU 0.00>`, memory is increased but gpu and cpu is decreased.

We should deal with both the increased and decreased value.

zionwu:

Fixes #729

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -111,11 +111,15 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		// If no queues, break
 		if totalWeight == 0 {
+			glog.V(4).Infof("Exiting when total weight is 0")
 			break
 		}
 
 		// Calculates the deserved of each Queue.
-		deserved := api.EmptyResource()
+		// increasedDeserved is the increased value for attr.deserved of processed queues
+		// decreasedDeserved is the decreased value for attr.deserved of processed queues
+		increasedDeserved := api.EmptyResource()
+		decreasedDeserved := api.EmptyResource()
 		for _, attr := range pp.queueOpts {
 			glog.V(4).Infof("Considering Queue <%s>: weight <%d>, total weight <%d>.",
 				attr.name, attr.weight, totalWeight)
@@ -125,20 +129,26 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 
 			oldDeserved := attr.deserved.Clone()
 			attr.deserved.Add(remaining.Clone().Multi(float64(attr.weight) / float64(totalWeight)))
-			if !attr.deserved.LessEqual(attr.request) {
+
+			if attr.request.Less(attr.deserved) {
 				attr.deserved = helpers.Min(attr.deserved, attr.request)
 				meet[attr.queueID] = struct{}{}
+				glog.V(4).Infof("queue <%s> is meet", attr.name)
+
 			}
 			pp.updateShare(attr)
 
 			glog.V(4).Infof("The attributes of queue <%s> in proportion: deserved <%v>, allocate <%v>, request <%v>, share <%0.2f>",
 				attr.name, attr.deserved, attr.allocated, attr.request, attr.share)
 
-			deserved.Add(attr.deserved.Clone().Sub(oldDeserved))
+			increased, decreased := attr.deserved.Diff(oldDeserved)
+			increasedDeserved.Add(increased)
+			decreasedDeserved.Add(decreased)
 		}
 
-		remaining.Sub(deserved)
+		remaining.Sub(increasedDeserved).Add(decreasedDeserved)
 		if remaining.IsEmpty() {
+			glog.V(4).Infof("Exiting when remaining is empty:  <%v>", remaining)
 			break
 		}
 	}
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -276,6 +276,41 @@ func (r *Resource) LessEqual(rr *Resource) bool {
 	return true
 }
 
+// Diff calculate the difference between two resource
+func (r *Resource) Diff(rr *Resource) (*Resource, *Resource) {
+	increasedVal := EmptyResource()
+	decreasedVal := EmptyResource()
+	if r.MilliCPU > rr.MilliCPU {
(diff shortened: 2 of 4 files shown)


zionwu:
@hex108 @k82cn Could you please review this PR? Thanks.

k82cn:
is it possible to add an e2e test for that?

zionwu:
> is it possible to add an e2e test for that?

Let me spend some time to see if it is possible to add e2e test.

zionwu:
@k82cn,  I have added the e2e test "Proportion".

Without my fix, the test will fail. You can verify at this CI Test:  https://travis-ci.org/kubernetes-sigs/kube-batch/builds/535261793?utm_source=github_status&utm_medium=notification

With my fix, the test can pass, which you can check the latest CI result.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -125,20 +129,26 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 
 			oldDeserved := attr.deserved.Clone()
 			attr.deserved.Add(remaining.Clone().Multi(float64(attr.weight) / float64(totalWeight)))
-			if !attr.deserved.LessEqual(attr.request) {
+
+			if attr.request.Less(attr.deserved) {
```
hex108, line 133: There has already been some problems caused by `Less` and `LessEqual`. The problems will be solved If  the type `MilliCPU`, `Memory` and values of `ScalarResources` change from `float64` to `resource.Quantity` or `int64`. 

@k82cn  What is your option?

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -125,20 +129,26 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
...
-			deserved.Add(attr.deserved.Clone().Sub(oldDeserved))
+			increased, decreased := attr.deserved.Diff(oldDeserved)
+			increasedDeserved.Add(increased)
+			decreasedDeserved.Add(decreased)
 		}
 
-		remaining.Sub(deserved)
+		remaining.Sub(increasedDeserved).Add(decreasedDeserved)
 		if remaining.IsEmpty() {
+			glog.V(4).Infof("exiting when remaining is empty:  <%v>", remaining)
```
hex108: s/exiting/Exiting/

zionwu: Updated.

pkg/scheduler/plugins/proportion/proportion.go
k82cn: >  The problems will be solved If the type MilliCPU, Memory and values of ScalarResources change from float64 to resource.Quantity or int64.

Do you mean if we change to int64, this fix is not necessary?

hex108: I think so, because it is caused by `LessEqual` is not `<=` actually.

k82cn: > LessEqual is not <= actually.

Can you share some case for that?

zionwu: In my opinion, this issue still exist even with resource.Quantity or int64.
The problem is that Resource is consist of MilliCPU, Memory and ScalarResources, and `LessEqual` requires all 3 members to be lessEqual:
```golang
func (r *Resource) LessEqual(rr *Resource) bool {
	isLess := (r.MilliCPU < rr.MilliCPU || math.Abs(rr.MilliCPU-r.MilliCPU) < minMilliCPU) &&
		(r.Memory < rr.Memory || math.Abs(rr.Memory-r.Memory) < minMemory)
	if !isLess {
		return false
	}

	if r.ScalarResources == nil {
		return true
	}

	for rName, rQuant := range r.ScalarResources {
		if rr.ScalarResources == nil {
			return false
		}

		rrQuant := rr.ScalarResources[rName]
		if !(rQuant < rrQuant || math.Abs(rrQuant-rQuant) < minMilliScalarResources) {
			return false
		}
	}

	return true
}
```
`!attr.deserved.LessEqual(attr.request)` is true when only one member of  attr.deserved is larger than attr.request. But we need all 3 members of attr.deserved larger than attr.request. So we should always use `attr.request.Less(attr.deserved)`  no matter what type of the 3 member is.

k82cn:
LGTM overall, thanks 👍

pkg/scheduler/plugins/proportion/proportion.go
hex108: Agree with @zionwu , the error is because we use `attr.request.LessEqual(attr.deserved)` but we should use `attr.request.Less(attr.deserved)`.

hex108:
/lgtm

for this PR.

zionwu:
@k82cn  Could you approve this PR? Thanks.

PR #742, opened 2019-04-11
Fix Golint Errors in pkg/controllers/job package

thandayuthapani:
Fix Golint Errors in following Package
- pkg/controllers/job	
- pkg/controllers/job/helpers	
- pkg/controllers/job/plugins	
- pkg/controllers/job/plugins/env
- pkg/controllers/job/plugins/ssh	
- pkg/controllers/job/state


TommyLike:
Fantastic!

PR #750, opened 2019-04-12
[Add Job Resource] Support build&e2e tests workflow

TommyLike:
For supporting job resource in kube batch.


TommyLike:
@k82cn The workflow for new build and e2e test process has been added into our makefile, but it's not enabled in Travis since:
1. It's a very long process compared with what we have in now
2. Tested in local environment and it's not stable now.

Usage:
`make e2e-kind`

k82cn (commented):
/lgtm
/approve

k82cn:
> It's a very long process compared with what we have in now

What's different with what we have in volcano.sh/volcano repo?

> Tested in local environment and it's not stable now.

That's ok to handle it as bug fix.

PR #751, opened 2019-04-12
Add UT cases for Preempt Action

thandayuthapani:
Added UT Cases for Preempt Action

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt_test.go
+++ b/pkg/scheduler/actions/preempt/preempt_test.go
@@ -18,15 +18,185 @@ package preempt
 
 import (
 	"testing"
+	"time"
 
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+
+	kbv1 "github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
+	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api"
+	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/cache"
+	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/conf"
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/framework"
-
-	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/plugins/drf"
+	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/plugins/conformance"
+	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/plugins/gang"
+	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/util"
 )
 
 func TestPreempt(t *testing.T) {
-	framework.RegisterPluginBuilder("drf", drf.New)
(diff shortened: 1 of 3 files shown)


thandayuthapani:
Hi @TommyLike , @hzxuzhonghu  can you please review this PR?

pkg/scheduler/actions/preempt/preempt_test.go
```diff
@@ -18,15 +18,185 @@ package preempt
...
+	tests := []struct {
+		name      string
+		podGroups []*kbv1.PodGroup
+		pods      []*v1.Pod
+		nodes     []*v1.Node
+		queues    []*kbv1.Queue
+		expected  int
+	}{
+		{
+			name: "one Job with two Pods on one node",
```
hzxuzhonghu, line 51: Confused about the name,  I am not familiar with the scheduler part, though.

thandayuthapani: Those are just test case names.  Have given a one line description about what testcase is trying to do.

pkg/scheduler/util/test_utils.go
```diff
@@ -0,0 +1,163 @@
...
+		},
+		Status: v1.NodeStatus{
+			Capacity:    alloc,
+			Allocatable: alloc,
+		},
+	}
+}
+
+// BuildPod builts Pod object
+func BuildPod(ns, n, nn string, p v1.PodPhase, req v1.ResourceList, groupName string, labels map[string]string, selector map[string]string) *v1.Pod {
```
hzxuzhonghu: prefer `namespace, name, nodeName `

pkg/scheduler/actions/preempt/preempt_test.go
hzxuzhonghu: Yes, i know. Just curious why one job and where is the job? Overall just a nit.

pkg/scheduler/actions/preempt/preempt_test.go
```diff
@@ -18,15 +18,185 @@ package preempt
...
+			podGroups: []*kbv1.PodGroup{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "pg1",
+						Namespace: "c1",
+					},
+					Spec: kbv1.PodGroupSpec{
+						Queue: "c1",
```
hzxuzhonghu: nit: The queue name is `c1`  and the namespace is also called `c1`

pkg/scheduler/actions/preempt/preempt_test.go
thandayuthapani: One podGroup is created in that test case, so all pods in a PodGroup is handled as all tasks under a Job in code.  So here only one PodGroup is created, which is internally(in code) one job in that queue.  That is why termed the test case name as one job.

pkg/scheduler/util/test_utils.go
thandayuthapani: Done

pkg/scheduler/actions/preempt/preempt_test.go
thandayuthapani: Done.  Have changed queue name to q1

pkg/scheduler/actions/preempt/preempt_test.go
hzxuzhonghu: ok

PR #753, opened 2019-04-13
Enabled e2e-kind.

k82cn:

**Release note**:
```release-note
None
```


TommyLike:
It seems that the log files will be truncated, need to find out a way to store them in another place.

PR #772, opened 2019-04-15
Remove reclaim and preempt in deployment

### LINKED ISSUE #770: Removed reclaim & preempt actions in vk-scheduler (kube-batch)

/kind bug

**Description**:

Remove reclaim & preemption actions from volcano deployment:

https://github.com/kubernetes-sigs/volcano/blob/master/deployment/volcano/config/kube-batch.conf

thandayuthapani:
Remove Reclaim and Preempt from deployment Conf
**Which issue(s) this PR fixes**
Fixes https://github.com/kubernetes-sigs/volcano/issues/770

### DIFF
--- a/deployment/volcano/config/kube-batch-ci.conf
+++ b/deployment/volcano/config/kube-batch-ci.conf
@@ -1,4 +1,4 @@
-actions: "allocate, backfill"
+actions: "reclaim, allocate, backfill, preempt"
 tiers:
 - plugins:
   - name: priority
--- a/deployment/volcano/config/kube-batch.conf
+++ b/deployment/volcano/config/kube-batch.conf
@@ -1,4 +1,4 @@
-actions: "reclaim, allocate, backfill, preempt"
+actions: "allocate, backfill"
 tiers:
 - plugins:
   - name: priority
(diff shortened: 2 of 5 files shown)


k82cn:
we also need to update the configuration in deployment; and we need a copy for CI.

thandayuthapani:
Have made changes in deployment/volcano/config and the same is being is used by CI

k82cn:
We have reclaim/preemption test in CI; how can we pass those CI with depoloyments?

config/kube-batch-conf.yaml
```diff
@@ -1,4 +1,4 @@
-actions: "reclaim, allocate, backfill, preempt"
+actions: "allocate, backfill"
```
TommyLike, line 1: Why changing this file in config folder? Will we deprecate them both?

deployment/volcano/config/kube-batch-ci.conf
```diff
@@ -0,0 +1,11 @@
+actions: "reclaim, allocate, backfill, preempt"
```
TommyLike, line 1: I thought you were going to remove them in this file, not the original file. And if I am right, you need to update testcase as well.

thandayuthapani: Since in CI, we need to test every action, deployment used in CI is same as the deployment provided for user.  So, I have added new config file for CI which will be loaded only during CI run, when user is deploying using helm it will use default config file which will have allocate, backfill alone, which is our requirement as mentioned in the issue https://github.com/kubernetes-sigs/volcano/issues/770

PR #775, opened 2019-04-15
Implement error code handling in lifecyclePolicy

hzxuzhonghu:

Implement error code handling in lifecyclePolicy

**Release note**:

```release-note
Implement error code handling in lifecyclePolicy.
```

### DIFF
--- a/pkg/controllers/job/job_controller_handler.go
+++ b/pkg/controllers/job/job_controller_handler.go
@@ -212,9 +212,14 @@ func (cc *Controller) updatePod(oldObj, newObj interface{}) {
 	}
 
 	event := vkbatchv1.OutOfSyncEvent
+	var exitCode int32
 	if oldPod.Status.Phase != v1.PodFailed &&
 		newPod.Status.Phase == v1.PodFailed {
 		event = vkbatchv1.PodFailedEvent
+		// TODO: currently only one container pod is supported by volcano
+		// Once multi containers pod is supported, update accordingly.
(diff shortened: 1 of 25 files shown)


pkg/controllers/job/job_controller_handler.go
```diff
@@ -212,9 +212,14 @@ func (cc *Controller) updatePod(oldObj, newObj interface{}) {
 	}
 
 	event := vkbatchv1.OutOfSyncEvent
+	var exitCode int32
 	if oldPod.Status.Phase != v1.PodFailed &&
 		newPod.Status.Phase == v1.PodFailed {
 		event = vkbatchv1.PodFailedEvent
+		// TODO: currently only one container pod is supported by volcano
+		// Once multi containers pod is supported, update accordingly.
+		exitCode = newPod.Status.ContainerStatuses[0].State.Terminated.ExitCode
```
k82cn, line 221: let's open an issue to trace that. one option is to set something like "1:1" which means considering second container's exist code (or by container name, e.g. worker:-2)

k82cn: the default value is 0, the "first" container.

k82cn:
LGTM overall

pkg/controllers/job/job_controller_handler.go
hzxuzhonghu: I will open an issue following this.

PR #784, opened 2019-04-16
Add generate code tools.

TommyLike:
Add generate code tools and make command

**Release note**:

```release-note
None
```


hzxuzhonghu:
/lgtm

please fix the verify

Makefile
```diff
@@ -71,6 +71,9 @@ run-test:
 e2e: kube-batch
 	hack/run-e2e.sh
 
+generate-code:
```
hzxuzhonghu, line 74: I would suggest `generate-code: generate-deepcopy`, so developers can call only `make generate-code `

TommyLike: The function of ``generate-deepcopy`` is already been included into ``generate-code``

hzxuzhonghu: I noticed that, can i say the `generate-deepcopy` is a redundant or a legacy stuff, which can be removed?

PR #795, opened 2019-04-17
support extended resource for kube-batch

wackxu:
@k82cn 

**Release note**:

```release-note
support extended resource for kube-batch
```

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -21,20 +21,24 @@ import (
 	"math"
 
 	"k8s.io/api/core/v1"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 )
 
 // Resource struct defines all the resource type
 type Resource struct {
 	MilliCPU float64
 	Memory   float64
-	MilliGPU float64
+
+	// ScalarResources
+	ScalarResources map[v1.ResourceName]float64
+
 	// MaxTaskNum is only used by predicates; it should NOT
 	// be accounted in other operators, e.g. Add.
 	MaxTaskNum int
 }
 
 const (
-	//GPUResourceName need to follow https://github.com/NVIDIA/k8s-device-plugin/blob/66a35b71ac4b5cbfb04714678b548bd77e5ba719/server.go#L20
+	// GPUResourceName need to follow https://github.com/NVIDIA/k8s-device-plugin/blob/66a35b71ac4b5cbfb04714678b548bd77e5ba719/server.go#L20
 	GPUResourceName = "nvidia.com/gpu"
 )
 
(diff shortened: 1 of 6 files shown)


TommyLike:
This patch needs a rebase since the change of repo name

pkg/scheduler/api/resource_info.go
```diff
@@ -88,18 +111,27 @@ func (r *Resource) IsZero(rn v1.ResourceName) bool {
...
 		return r.Memory < minMemory
-	case GPUResourceName:
-		return r.MilliGPU < minMilliGPU
 	default:
-		panic("unknown resource")
+		if r.ScalarResources == nil {
+			return true
+		}
+
+		return r.ScalarResources[rn] < minMilliScalarResources
```
k82cn, line 123: check whether `rn` is valid.

k82cn:
please add some test for that.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -232,7 +232,7 @@ func (pp *proportionPlugin) updateShare(attr *queueAttr) {
 	res := float64(0)
 
 	// TODO(k82cn): how to handle fragment issues?
-	for _, rn := range api.ResourceNames() {
+	for _, rn := range attr.allocated.ResourceNames() {
```
hex108: `attr.allocated`'s extended resource is initialed to nil, so its `ResourceNames()`  will not contain extended resource?

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -160,7 +160,7 @@ func (drf *drfPlugin) updateShare(attr *drfAttr) {
 
 func (drf *drfPlugin) calculateShare(allocated, totalResource *api.Resource) float64 {
 	res := float64(0)
-	for _, rn := range api.ResourceNames() {
+	for _, rn := range allocated.ResourceNames() {
```
hex108: Might it be better to use `totalResource` because `allocated`'s `ScalarResources` perhaps is nil?

wackxu: Thanks for your review, updated

pkg/scheduler/plugins/proportion/proportion.go
wackxu: update to use attr.deserved

pkg/scheduler/api/resource_info.go
wackxu: Done

wackxu:
Add some UT, @k82cn @hex108 @TommyLike PTAL

hex108:
/lgtm

Thanks!

PR #805, opened 2019-04-19
Added backfill 

pdgetrf:
Added backfill action that allows lower priority jobs to run when higher priority jobs cannot run but still hold resources

Tested with added unit tests, e2e tests and local manual test


pdgetrf:
signed CLA

config/kube-batch-conf.yaml
```diff
@@ -9,3 +9,4 @@ tiers:
   - name: predicates
   - name: proportion
   - name: nodeorder
+enable-backfill: true
```
k82cn: have we addressed my previous comments in this PR?

k82cn:
btw, please also open a PR on design doc.

k82cn:
and please also append the review comments last time and highlight how those comments were addressed.

dhzhuo:
@k82cn  Design doc of backfill and starvation prevention can be found in [PR 821](https://github.com/kubernetes-sigs/kube-batch/pull/821).

config/kube-batch-conf.yaml
dhzhuo: This PR addressed some key issues you raised earlier. The other less urgent issues will be addressed later after this PR.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +70,100 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
 		}
 	}
+
+	if ssn.ActionOptions == nil {
+		return
+	}
+
+	backfillEnabled, err := strconv.ParseBool(ssn.ActionOptions[conf.BackfillFlagName])
+	if err != nil {
+		return
```
dhzhuo: Add a log statement here?

pkg/scheduler/conf/scheduler_conf.go
```diff
@@ -19,16 +19,25 @@ package conf
...
 type SchedulerConfiguration struct {
 	// Actions defines the actions list of scheduler in order
-	Actions string `yaml:"actions"`
+	Actions []SchedulerAction `yaml:"actions"`
 	// Tiers defines plugins in different tiers
 	Tiers []Tier `yaml:"tiers"`
 }
 
+// BackfillFlagName the flag to enable backfill for non-best effort jobs
+const BackfillFlagName = "enabledNonBestEffort"
```
dhzhuo: How about name it as `backfillNonBestEffortJobs`?

pkg/scheduler/api/types.go
```diff
@@ -53,10 +59,40 @@ const (
...
+	// #(Allocated Tasks) >= Job.MinAvailable
+	// A Ready job can be dispatch to a node right away.
+	Ready JobReadiness = 1 << iota
+
+	// OverResourceReady : a job is OverResourceReady if the job is not Ready for dispatch, but
+	// the number of tasks in Allocated state exceeds the job's minim task
+	// number requirement. In other words,
+	// #(Allocated Tasks) < Job.MinAvailable &&
+	// #(Allocated Tasks) + #(AllocatedOverBackFill Tasks) >= Job.MinAvailable
+	OverResourceReady
```
dhzhuo: I can live with this name. But personally I still prefer `ConditionallyReady`. :-)

dhzhuo (commented):
Looks good. Just have a few minor suggestion.

doc/design/plugin-conf.md
```diff
@@ -22,7 +22,14 @@ scenario on the fly.
 The following YAML format will be introduced for dynamic plugin configuration:
 
 ```yaml
-actions: "list_of_action_in_order"
+actions:
+- name: action_1
+  options:
+    key1: val1
```
k82cn, line 28: refer to plugin's arguments.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -18,6 +18,8 @@ package backfill
 
 import (
 	"github.com/golang/glog"
+	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/conf"
```
k82cn, line 21:
```go
sys pkg

third part pkg

kube-batch pkg
```

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +70,100 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
 			}
 		}
 	}
+
+	if ssn.ActionOptions == nil {
+		return
+	}
+
+	backfillEnabled, err := strconv.ParseBool(ssn.ActionOptions[conf.BackfillFlagName])
```
k82cn: refer to plugin's argument on how to get value.

pkg/scheduler/actions/backfill/backfill.go
hex108: To solve this kind of problem, we could use [goimports](http://goinbigdata.com/goimports-vs-gofmt/) to do format. How about it?

pkg/scheduler/api/job_info.go
```diff
@@ -18,7 +18,9 @@ package api
 
 import (
 	"fmt"
+	"github.com/golang/glog"
```
k82cn: ditto

pkg/scheduler/api/job_info.go
```diff
@@ -51,6 +53,10 @@ type TaskInfo struct {
 	VolumeReady bool
 
 	Pod *v1.Pod
+
+	// Indicates whether or not the task is a backfill task. Either persist
+	// across sessions, or add this flag in pod annotation.
+	IsBackfill bool
```
k82cn: prefer to use slice, e.g. []Condition, Backfill is one of condition in running status.

pkg/scheduler/api/job_info.go
```diff
@@ -65,6 +71,32 @@ func getJobID(pod *v1.Pod) JobID {
...
+			hasBackfillAnnotation, err := strconv.ParseBool(val)
+			if err != nil {
+				glog.Errorf("Invalid backfill annotation value '%s': %s", pod.Annotations[v1alpha1.BackfillAnnotationKey], err)
+				return false
+			}
+			glog.Infof("Restored backfill status for pod %s", pod.Name)
+
+			isPodScheduled := false
+			if pod.Status.Conditions != nil {
+				for _, cond := range pod.Status.Conditions {
```
k82cn: check nodename should be enough.

pkg/scheduler/api/job_info.go
```diff
@@ -65,6 +71,32 @@ func getJobID(pod *v1.Pod) JobID {
 	return ""
 }
 
+// CheckBackfill check if the pod has valid backfill annotation
+func CheckBackfill(pod *v1.Pod) bool {
```
k82cn: is anyone else using this helper func? if not, should be private helper.

pkg/scheduler/api/resource_info.go
```diff
@@ -141,22 +141,17 @@ func (r *Resource) Add(rr *Resource) *Resource {
...
+	for rrName, rrQuant := range rr.ScalarResources {
+		if r.ScalarResources == nil {
+			return r
 		}
-
-		return r
+		r.ScalarResources[rrName] -= rrQuant
 	}
 
-	panic(fmt.Errorf("Resource is not sufficient to do operation: <%v> sub <%v>",
```
k82cn, line 158: keep panic

pkg/scheduler/api/test_utils.go
```diff
@@ -18,6 +18,7 @@ package api
 
 import (
 	"fmt"
+	"github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
```
k82cn: ditto

pkg/scheduler/api/types.go
```diff
@@ -23,6 +23,12 @@ const (
 	// Pending means the task is pending in the apiserver.
 	Pending TaskStatus = 1 << iota
 
+	// AllocatedOverBackfill means that the task is allocated with resources
+	// that are currently occupied by backfill tasks.
+	// Task T is allocated to Node N as an AllocatedOverBackfill task if,
+	// and only if, N.IdleResource < T.RequestResource <= N.AllocatableResource.
+	AllocatedOverBackfill
```
k82cn: Can we reuse `Pipelined`? If not, prefer `Waiting` which means waiting for running task to release resources.

pkg/scheduler/actions/backfill/backfill.go
k82cn: that's ok to me :)

pkg/scheduler/api/types.go
```diff
@@ -53,10 +59,40 @@ const (
...
 
+// JobReadiness type of job readiness
+type JobReadiness int
+
+const (
+	// Ready : a job is Ready if the number of tasks in Allocated state
+	// exceeds the job's minimum task number requirement. In other words,
+	// #(Allocated Tasks) >= Job.MinAvailable
+	// A Ready job can be dispatch to a node right away.
+	Ready JobReadiness = 1 << iota
```
k82cn, line 75: we dependent on plugin to decide whether a job is ready, refer to `plugin/gang` & `JobReady` callback.

pkg/scheduler/scheduler.go
```diff
@@ -96,6 +104,7 @@ func (pc *Scheduler) runOnce() {
 
 	for _, action := range pc.actions {
 		actionStartTime := time.Now()
+		ssn.ActionOptions = pc.actionOptions[action.Name()]
```
k82cn: why not add options to action struct?

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -44,6 +44,20 @@ func (gp *gangPlugin) Name() string {
 	return "gang"
 }
 
+func backFillEligible(obj interface{}) bool {
+	job := obj.(*api.JobInfo)
+
+	allPending := true
+	for _, task := range job.Tasks {
+		if task.Status != api.Pending {
```
k82cn, line 52: if some tasks are `Allocated`, we can not backfill it?

k82cn: is there any callback on select backfill candidate? We discussed one case that only the job with timeout can reuse the idle resources. And the timeout threshold is configurable.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: added

pkg/scheduler/conf/scheduler_conf.go
pdgetrf: do you mean the variable name or the string?

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: the new commit in the master branch to use goimports brought lots of lines like the this:

image

here v1 is redundant. Why exactly?

pkg/scheduler/conf/scheduler_conf.go
dhzhuo: I was referring to rename `enableNonBestEffort` to `backfillNonBestEffortJobs`.

pkg/scheduler/api/types.go
pdgetrf: AllocatedOverBackfill cannot be replaced with Pipelined. Tasks allocated with AllocatedOverBackfill status will release their resources during the backfill action if their belonging jobs are still not ready. We do not do the same for pipelined task.

pdgetrf: On the other hand, "Waiting" does not have a clear meaning on what's being waited for. What about "AllocatedWithBackfilledResource". It's long but I'd trade length for confusion.

pkg/scheduler/scheduler.go
pdgetrf: Action is an interface

pkg/scheduler/api/types.go
pdgetrf: fixed

pkg/scheduler/plugins/gang/gang.go
pdgetrf: yes only job with all pending tasks will be backfilling candidate. this possibly can be relaxed later. 

for the second question, it look a lot like what we have for starvation in the other PR following this one.

pkg/scheduler/api/resource_info.go
pdgetrf: by design negative resource is now possible when resource is allocated using backfill job resources.

pkg/scheduler/conf/scheduler_conf.go
pdgetrf: fixed

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: working on this one.

pkg/scheduler/api/job_info.go
pdgetrf: will fix after the goimports question.

pkg/scheduler/actions/backfill/backfill.go
k82cn: that's ok to me if generated by tools; please verify that.

pkg/scheduler/api/resource_info.go
k82cn: that's in-correct to me; I think we discussed this before, we need to keep panic; if negative integer is necessary, add other field to handle that.

pkg/scheduler/api/types.go
k82cn: I do not want to add a phase for just one feature; we should add phase for general purpose.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: Seems to me more like a bug from the tools.

@hex108 made the change. Could you please verify these are generated on purpose?

hex108: It is intended, https://github.com/golang/go/issues/30051

pkg/scheduler/scheduler.go
k82cn: add a member func to get options.

pkg/scheduler/plugins/gang/gang.go
k82cn: > for the second question, it should a lot like what we have for starvation in the other PR following this one.

This 's not related to reseravation to me; so why handle it in other pr?

doc/design/plugin-conf.md
pdgetrf: fixed

pkg/scheduler/scheduler.go
pdgetrf: fixed

pkg/scheduler/api/job_info.go
pdgetrf: fixed

pkg/scheduler/api/resource_info.go
pdgetrf: "if negative integer is necessary" then we should allow it. why bother with another field?

pkg/scheduler/api/types.go
pdgetrf: renamed to AllocatedPendingResource

k82cn: > renamed to AllocatedPendingResource

Try to use one word, similar to other status.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: will rebase nonetheless. my question remains.

pkg/scheduler/plugins/gang/gang.go
pdgetrf: backfill does not have any timeout configuration. only starvation does.

pkg/scheduler/api/types.go
pdgetrf: One word will cause confusion and why do we prefer that?

pkg/scheduler/plugins/gang/gang.go
dhzhuo: > We discussed one case that only the job with timeout can reuse the idle resources. And the timeout threshold is configurable.

@k82cn  Can you please elaborate the above use case?

pkg/scheduler/api/types.go
pdgetrf: the closest I can think of is "Expecting"

pdgetrf: renamed to "Borrowing"

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +73,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
 			}
 		}
 	}
+
+	if len(alloc.arguments) == 0 {
+		return
+	}
+
+	backfillEnabled, err := strconv.ParseBool(alloc.arguments[conf.BackfillFlagName])
```
hex108: Maybe it could initialized in function `New` at line 33? `backfillEnabled` is a little confusing because it is actually for enable backfill for non best effort task.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +73,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+		}
+
+		// Release resources allocated to unready top dog jobs so that
+		// we can back fill more jobs in the next step.
+		resourceReleased := false
+		for _, job := range ssn.Jobs {
+			if ssn.JobReady(job) || ssn.JobBackfillReady(job) {
+				glog.V(3).Infof("Disable backfill on job <%v/%v>", job.Namespace, job.Name)
+			} else {
+				releaseReservedResources(ssn, job)
```
hex108, line 100: How about `resourceReleased = releaseReservedResources(ssn, job)` and add a return value for the function? According to the function, there is some case that it could not release any resource.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +73,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+			if err := ssn.PredicateFn(task, node); err != nil {
+				glog.V(3).Infof("Predicates failed for task <%s/%s> on node <%s>: %v",
+					task.Namespace, task.Name, node.Name, err)
+				continue
+			}
+
+			if task.InitResreq.LessEqual(node.Idle) {
+				task.Condition.IsBackfill = true
+				glog.V(3).Infof("Binding backfill task <%v/%v> to node <%v>", task.Namespace, task.Name, node.Name)
+				if err := ssn.Allocate(task, node); err != nil {
```
hex108, line 150: According to the backfill doc in https://github.com/kubernetes-sigs/kube-batch/pull/821, it will not dispatch tasks in this phase, but `ssn.Alocate` will dispatch tasks if job is ready.

pkg/scheduler/api/types.go
```diff
@@ -23,6 +23,12 @@ const (
 	// Pending means the task is pending in the apiserver.
 	Pending TaskStatus = 1 << iota
 
+	// Borrowing means that the task is allocated borrowing resources
+	// that are currently occupied by tasks.
+	// Task T is allocated to Node N as an Borrowing task if,
+	// and only if, N.IdleResource < T.RequestResource <= N.AllocatableResource.
+	Borrowing
```
k82cn, line 30: will that "Borrowing" resources will be return?

deployment/volcano/config/kube-batch-ci.conf
```diff
@@ -1,4 +1,10 @@
-actions: "reclaim, allocate, backfill, preempt"
+actions:
+- name: reclaim
+- name: allocate
+- name: backfill
+  options:
```
k82cn: `options` is not necessary.

pkg/scheduler/actions/backfill/backfill.go
k82cn: refer to `GetInt` in `Arguments`.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: I don't see the need

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: no it will not. please read what's in ssn.Alocate

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: renamed to backfillNonBestEffortJobsEnabled. clear enough?

deployment/volcano/config/kube-batch-ci.conf
pdgetrf: not relevant any more. rebased away.

doc/design/plugin-conf.md
```diff
@@ -22,7 +22,14 @@ scenario on the fly.
 The following YAML format will be introduced for dynamic plugin configuration:
 
 ```yaml
-actions: "list_of_action_in_order"
+actions:
+- name: action_1
+  options:
```
k82cn, line 27: ~~remove `options`, refer to plugin arguments.~~

just check the code of plugin, that's ok to me right now.

pkg/apis/scheduling/v1alpha1/labels.go
```diff
@@ -19,3 +19,5 @@ package v1alpha1
 // GroupNameAnnotationKey is the annotation key of Pod to identify
 // which PodGroup it belongs to.
 const GroupNameAnnotationKey = "scheduling.k8s.io/group-name"
+
+const BackfillAnnotationKey = "scheduling.k8s.io/kube-batch-backfill"
```
k82cn: just `backfill-task` or something else, do not mention kube-batch here.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
 			}
 		}
 	}
+
+	if len(alloc.arguments) == 0 {
+		return
+	}
+
+	backfillNonBestEffortJobsEnabled, err := strconv.ParseBool(alloc.arguments[conf.BackfillFlagName])
```
k82cn: refer to `GetInt` in Arguments for that.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+			if !ssn.BackFillEligible(job) {
+				continue
+			}
+			backFillCandidates = append(backFillCandidates, job)
+		}
+
+		// Release resources allocated to unready top dog jobs so that
+		// we can back fill more jobs in the next step.
+		resourceReleased := false
+		for _, job := range ssn.Jobs {
```
k82cn, line 96: can we merge this with above loop? If job can not be backfill by other jobs, why should it release allocated resources?

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+		if resourceReleased {
+			for _, job := range backFillCandidates {
+				backFill(ssn, job)
+			}
+		}
+	}
+}
+
+// Releases resources allocated to the given job back to the cluster.
+func releaseReservedResources(ssn *framework.Session, job *api.JobInfo) {
```
k82cn, line 114: Prefer to add this func to session as `Deallocate`.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+		}
+		if allocateFailed {
+			glog.V(3).Infof("Aborted backilling job %s", job.Name)
+			break
+		}
+	}
+
+	if !ssn.JobReady(job) {
+		glog.V(3).Infof("Job <%v/%v> is not ready. Release its resources.", job.Namespace, job.Name)
+		releaseReservedResources(ssn, job)
```
k82cn, line 165: please enhancement `Statement` in framework for that, refer to preempt action.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+				glog.Errorf("Failed to remove task %v from node %v: %s", task.Name, node.Name, err)
+				continue
+			}
+
+			glog.V(4).Infof("Removed task %s from node %s. Idle: %+v; Used: %v; Releasing: %v.", task.Name, node.Name, node.Idle, node.Used, node.Releasing)
+		}
+	}
+}
+
+func backFill(ssn *framework.Session, job *api.JobInfo) {
```
k82cn, line 135: this should be also a func in session.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+	glog.V(3).Infof("Backfill job <%v/%v>", job.Namespace, job.Name)
+
+	for _, task := range job.TaskStatusIndex[api.Pending] {
+		allocateFailed := false
+		for _, node := range ssn.Nodes {
```
k82cn, line 140: how about nodeOrder?

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+			if err := ssn.PredicateFn(task, node); err != nil {
+				glog.V(3).Infof("Predicates failed for task <%s/%s> on node <%s>: %v",
+					task.Namespace, task.Name, node.Name, err)
+				continue
+			}
+
+			if task.InitResreq.LessEqual(node.Idle) {
+				task.Condition.IsBackfill = true
```
k82cn, line 148: `task.Condition = Backfill` or `task.Conditions=append(task.Conditions, Backfill)`

pkg/scheduler/actions/backfill/backfill.go
k82cn: oh, this func should not in session.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -68,6 +74,101 @@ func (alloc *backfillAction) Execute(ssn *framework.Session) {
...
+				glog.V(3).Infof("Binding backfill task <%v/%v> to node <%v>", task.Namespace, task.Name, node.Name)
+				if err := ssn.Allocate(task, node); err != nil {
```
k82cn, line 150: we should use `Backfill` for such case to avoid reuse `Allocate`.

pkg/scheduler/actions/backfill/backfill.go
k82cn: `task.Conditions=append(task.Conditions, Backfill)` is better to me.

pdgetrf: neither.

Backfill is one of the conditions. But not all conditions are boolean. Type should be flexible. Future new condition will be added to the Condition struct.

k82cn: we should use int for condition instead of bool which is not friendly enough to extend.

pkg/scheduler/api/types.go
pdgetrf: see, this is exactly why I didn't want to change to single word name. CONFUSION!!

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: why?

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: struct type is design for this kind of purpose. why not use it?

pkg/scheduler/api/job_info.go
```diff
@@ -371,6 +402,22 @@ func (ji *JobInfo) FitError() string {
...
+// GetBackillReadiness checks whether the job is ready or ready
+// using backill job's resources ConditionallyReady means the
+// job is ready but some resource is used by backfilled job
+func (ji *JobInfo) BackfillReady() bool {
+	allocatedTasks := ji.GetTasks(AllocatedStatuses()...)
+	allocatedTasksCnt := int32(len(allocatedTasks))
+
+	allocatedOverBackfillTasks := ji.GetTasks(Borrowing)
+	allocatedOverBackfillTasksCnt := int32(len(allocatedOverBackfillTasks))
+	if allocatedTasksCnt+allocatedOverBackfillTasksCnt >= ji.MinAvailable {
```
k82cn, line 413: we should not check `minAvailable` in Job, `minAvailable` is used by gang plugin. I need delegate to gang plugin to do that. So when we disable gang plugin, minAvailable will not work, refer to `JobReady`.

pkg/scheduler/api/node_info.go
```diff
@@ -74,9 +80,16 @@ func NewNodeInfo(node *v1.Node) *NodeInfo {
 // Clone used to clone nodeInfo Object
 func (ni *NodeInfo) Clone() *NodeInfo {
 	res := NewNodeInfo(ni.Node)
+	glog.V(3).Infof("new node <%v>: capability %v,  allocatable %v, idle %v, used %v, backfilled %v, releasing %v", ni.Name, ni.Capability.MilliCPU,
```
k82cn: please remove log in `Clone`.

pkg/scheduler/api/node_info.go
```diff
@@ -124,6 +141,12 @@ func (ni *NodeInfo) AddTask(task *TaskInfo) error {
 			ni.Idle.Sub(ti.Resreq)
 		}
 
+		// removing pod resources not maintained by kube-batch from node.ALlocatable
+		if ti.Job == "" && ti.Pod.Status.Phase == v1.PodRunning {
```
k82cn, line 145: Why change allocatable? It's reported from `Node`, not related to pod.

pkg/scheduler/api/node_info.go
```diff
@@ -193,3 +220,11 @@ func (ni *NodeInfo) Pods() (pods []*v1.Pod) {
 
 	return
 }
+
+// GetAccessibleResource get accessible resource which equals idle + backfilled
+func (ni *NodeInfo) GetAccessibleResource() *Resource {
+	idle := ni.Idle.Clone()
+	accessible := idle.Add(ni.Backfilled).Clone()
```
k82cn: It's not necessary to clone again, right?

pkg/scheduler/api/resource_info.go
k82cn: keep panic, it catches several issue there.

k82cn: btw, please get confirm before resolve conversation.

pkg/scheduler/api/types.go
k82cn: how did we fix that? I mean we should not have such kind of info in framework, framework is hard to decide whether a 'job' is ready, it all dependent on plugin to make decision. As we're going to support different type of workload.

pkg/scheduler/conf/scheduler_conf.go
```diff
@@ -37,6 +46,8 @@ type PluginOption struct {
 	EnabledJobOrder *bool `yaml:"enableJobOrder"`
 	// EnabledJobReady defines whether jobReadyFn is enabled
 	EnabledJobReady *bool `yaml:"enableJobReady"`
+	// EnabledJobBackfillReady defines whether jobBackfillReadyFn is enabled
+	EnabledJobBackfillReady *bool `yaml:"enableJobBackfillReady"`
```
k82cn, line 50: BackfillReady or BackfillEligable?

doc/design/plugin-conf.md
pdgetrf: type PluginOption struct {
	// The name of Plugin
	Name string `yaml:"name"`
        .......
	// Arguments defines the different arguments that can be given to different plugins
	Arguments map[string]string `yaml:"arguments"`
}

if there's plugin arguments, it will be:

- name: xxx
   arguments: 

I can rename option to arguments.

pkg/scheduler/api/node_info.go
pdgetrf: why?

pkg/scheduler/framework/session.go
```diff
@@ -234,15 +244,23 @@ func (ssn *Session) Pipeline(task *api.TaskInfo, hostname string) error {
 }
 
 //Allocate the task to the node in the session
-func (ssn *Session) Allocate(task *api.TaskInfo, hostname string) error {
+func (ssn *Session) Allocate(task *api.TaskInfo, node *api.NodeInfo) error {
```
k82cn, line 247: it is better to have `Backfill` func for that.

pkg/scheduler/api/node_info.go
pdgetrf: because as is the variable allocatable here contains resource already used by other pods not managed by kube-batch. they are literally not "allocable" by kube-batch and needs to be subtracted from allocable. we have seen starvation e2e test failure caused by this.

pdgetrf:
rebase is done.

pkg/scheduler/framework/session.go
```diff
@@ -279,6 +297,46 @@ func (ssn *Session) Allocate(task *api.TaskInfo, hostname string) error {
...
+
+		for _, pendingTask := range job.TaskStatusIndex[api.Allocated] {
+			if pendingTask.Condition.IsBackfill {
+
+				wg.Add(1)
+				go func(pendingTask *api.TaskInfo) {
+					defer wg.Done()
+
+					annotation := map[string]string{v1alpha1.BackfillAnnotationKey: "true"}
+					err := ssn.PatchAnnotation(pendingTask, annotation)
```
k82cn, line 314: The annotation should be also cached, there maybe out of sync between scheduler cache & apiserver.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -177,6 +187,27 @@ func (ssn *Session) Overused(queue *api.QueueInfo) bool {
 	return false
 }
 
+// JobReady invoke jobready function of the plugins
```
k82cn: s/JobReady/JobBackfillReady/g

pkg/scheduler/framework/session_plugins.go
```diff
@@ -198,6 +229,24 @@ func (ssn *Session) JobReady(obj interface{}) bool {
 	return true
 }
 
+// BackFillEligible check backfill eligiblity
+func (ssn *Session) BackFillEligible(obj interface{}) bool {
+	for _, tier := range ssn.Tiers {
+		for _, plugin := range tier.Plugins {
+			jrf, found := ssn.backFillEligibleFns[plugin.Name]
```
k82cn, line 236: should be also configurable

pkg/scheduler/plugins/gang/gang.go
k82cn: we need to make sure we have a callback to select which job can be backfilled to improve resource utilization. The job with timeout is a candidate.

k82cn: > yes only job with all pending tasks will be backfilling candidate. this possibly can be relaxed later.

In backfill action, we check eligible firstly before release allocated resources; so if one job has some allocated resource but can not start, this eligible callback will return false, right?

pkg/scheduler/scheduler.go
```diff
@@ -36,9 +36,11 @@ type Scheduler struct {
 	cache          schedcache.Cache
 	config         *rest.Config
 	actions        []framework.Action
+	actionOptions  map[string]map[string]string
```
k82cn: this options should be included in Action instead of scheduler.

pkg/scheduler/scheduler.go
```diff
@@ -36,9 +36,11 @@ type Scheduler struct {
...
 	plugins        []conf.Tier
 	schedulerConf  string
 	schedulePeriod time.Duration
+	enableBackfill bool
```
k82cn: action level option, not scheduler.

pkg/scheduler/util.go
```diff
@@ -19,17 +19,17 @@ package scheduler
...
 	"fmt"
 	"io/ioutil"
-	"strings"
-
-	"gopkg.in/yaml.v2"
 
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/conf"
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/framework"
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/plugins"
+	"gopkg.in/yaml.v2"
```
k82cn: keep the original format.

pkg/scheduler/util.go
```diff
@@ -60,16 +58,41 @@ func loadSchedulerConf(confStr string) ([]framework.Action, []conf.Tier, error)
...
+		if _, found := framework.GetAction(action.Name); !found {
+			return nil, fmt.Errorf("failed to found Action %s, ignore it", action.Name)
+		}
+
+		if action.Options == nil {
+			schedulerConf.Actions[i].Options = map[string]string{}
+		}
+
+		// set default value for backfill enabled
+		if action.Name == "backfill" {
```
k82cn: we should not check action name in framework.

pkg/scheduler/util.go
```diff
@@ -60,16 +58,41 @@ func loadSchedulerConf(confStr string) ([]framework.Action, []conf.Tier, error)
...
-			actions = append(actions, action)
-		} else {
-			return nil, nil, fmt.Errorf("failed to found Action %s, ignore it", actionName)
+	for i, action := range schedulerConf.Actions {
+		if _, found := framework.GetAction(action.Name); !found {
+			return nil, fmt.Errorf("failed to found Action %s, ignore it", action.Name)
+		}
+
+		if action.Options == nil {
+			schedulerConf.Actions[i].Options = map[string]string{}
```
k82cn: let's move options to each action accordingly.

doc/design/plugin-conf.md
pdgetrf: renamed to "arguments"

pkg/apis/scheduling/v1alpha1/labels.go
pdgetrf: fixed

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: there's some issue when merging this two loops. looking into it. 

if a job is not ready even if by borrowing resources from backfilled jobs, it will release its allocated resources  because it won't be ready anyway.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: no. this function needs to be called before backfill()

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: I thought we agree to not use preempt in this PR.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: will be addressed along with the starvation PR

pkg/scheduler/api/node_info.go
pdgetrf: fixed

pkg/scheduler/conf/scheduler_conf.go
pdgetrf: BackfillEligable checks whether a job can be backfilled (e.g. all task pending)

EnabledJobBackfillReady is a flag similar to EnabledJobReady in the master branch

pkg/scheduler/framework/session_plugins.go
pdgetrf: this is already fixed

pkg/scheduler/scheduler.go
pdgetrf: this is not used any more. removed.

pkg/scheduler/framework/session_plugins.go
pdgetrf: "configurable" as in "can be configured to turn on and off"?

pkg/scheduler/scheduler.go
pdgetrf: fixed

pkg/scheduler/util.go
pdgetrf: what's original format?

pkg/scheduler/util.go
pdgetrf: action.Options means "the Options of this action"

pkg/scheduler/util.go
pdgetrf: removed this name check

pdgetrf:
still working on a few issues. the e2e tests from the previous rollback in the master branch (not related to this PR) wrecked a mess. attempting to fix it now at least for backfill.

pkg/scheduler/actions/backfill/backfill.go
k82cn: move this func as a member of session, and renamed it to `Deallocate`. So we can reuse it later.

pkg/scheduler/actions/backfill/backfill.go
k82cn: I think we already sync up about this several times :( , move the fix into this PR.

pkg/scheduler/actions/backfill/backfill.go
k82cn: similar to `Pipeline` function; it's better to different func for different purpose.

pkg/scheduler/actions/backfill/backfill.go
k82cn: it's not about premption, it's about how preemption using `Statement`, please check the related code.

pkg/scheduler/actions/backfill/backfill.go
dhzhuo: This is backfill. Backfill does not require stable node order. The node order does not matter here.

pkg/scheduler/api/node_info.go
k82cn: we need to include all running resource in kube-batch.

pkg/scheduler/api/types.go
k82cn: Single word is better to me, but `Borrowing` and the comments does not describe it well.

pkg/scheduler/conf/scheduler_conf.go
k82cn: if so, why only provide one configuration?

pkg/scheduler/framework/session.go
```diff
@@ -279,6 +297,46 @@ func (ssn *Session) Allocate(task *api.TaskInfo, hostname string) error {
...
+			wg.Wait()
+			close(finished)
+		}()
+
+		select {
+		case <-finished:
+		case err := <-errChannel:
+			if err != nil {
+				fmt.Println("error ", err)
+				return err
```
k82cn, line 335: What happen if Pod is patched but did not dispatch because of other error?

pkg/scheduler/util.go
k82cn:
```
	"io/ioutil"

	"gopkg.in/yaml.v2"

	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/conf"
```

pkg/scheduler/util.go
```diff
@@ -60,16 +58,39 @@ func loadSchedulerConf(confStr string) ([]framework.Action, []conf.Tier, error)
...
+		if _, found := framework.GetAction(action.Name); !found {
+			return nil, fmt.Errorf("failed to found Action %s, ignore it", action.Name)
+		}
+
+		if action.Arguments == nil {
+			schedulerConf.Actions[i].Arguments = map[string]string{}
+		}
+
+		// set default value for backfill enabled
+		if _, found := schedulerConf.Actions[i].Arguments[conf.BackfillFlagName]; !found {
```
k82cn: It's not noly about the string, framework should not know anything about action. refer to plugins.

pkg/scheduler/framework/session.go
dhzhuo: If dispatch fails, the pod remains unbound in the current scheduling round, and therefore will not be started. In the next scheduling round, when we create TaskInfo for a Pod in `NewTaskInfo`, we will noticed that the Pod is unbound but has backfill annotation. In this case, we will remove backfill annotation from the Pod. 

@pdgetrf It looks to me that we do not yet have the logic to remove backfill annotation in `NewTaskInfo`.

pkg/scheduler/api/node_info.go
dhzhuo: Before we add backfill capability in this PR, kube-batch makes scheduling decisions by comparing a task's request resource amount against a node's *idle* resource amount. It does not use a node's *allocatable* resource amount. Kube-batch will be happy as long as the node's idle resource amount is correct. 

With the backfill capability added in this PR, kube-batch not only needs to check a node's idle resource amount, but also needs to check a node's *allocatable* amount.  For example, kube-batch may choose to backfill a task when the task's request resource amount is greater than a node's idle resource amount but less than a node's allocatable resource amount.  Therefore, kube-batch has to make sure both *idle* amount and *allocatable* amount are accurate. 

In a cluster where kube-batch scheduler runs side by side with other schedulers, a node may have multiple running Pods among which some are managed by kube-batch and some are managed by other schedulers.  Kube-batch can only preempt pods managed by itself. It has no right to preempt pods managed by other schedulers. In this case, we'd better deduct a node's allocatable resource amount when we see a running pod which does not belong to kube-batch. Otherwise, the node's allocatable resource amount will be bigger than it supposed to be.

pkg/scheduler/actions/backfill/backfill.go
k82cn: we need to call `nodeOrderFn` to get the "best" resources for it.

pkg/scheduler/api/node_info.go
k82cn: > Kube-batch can only preempt pods managed by itself. It has no right to preempt pods managed by other schedulers. 

In this case, we need to mark it as un-preemptable in conformance plugin.
As we discussed before, we should not change the meaning of this field.

pkg/scheduler/framework/session.go
k82cn: >  In this case, we will remove backfill annotation from the Pod.

It seems not implemented in this PR, right?

pkg/scheduler/framework/session_plugins.go
k82cn: yes.

pkg/scheduler/actions/backfill/backfill.go
dhzhuo: @k82cn  You are right. We should use `util.PredicateNodes`, `util.PrioritizeNodes`, and `util.SelectBestNode` as we do in allocate.go.

pdgetrf: > I think we already sync up about this several times :( , move the fix into this PR.

Same thing with starvation PR is based on the backfill PR which needs to be merged first. 

Same thing with backfill PR is a big PR and saying that it should be split into smaller PRs does not help. 

Same thing with we need to prioritize comments

Yeah, we have sync'd on these too so many times and look where we are now.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: backfill is doing allocation so we call allocate.

pkg/scheduler/api/node_info.go
pdgetrf: > > Kube-batch can only preempt pods managed by itself. It has no right to preempt pods managed by other schedulers.
> 
> In this case, we need to mark it as un-preemptable in conformance plugin.
> As we discussed before, we should not change the meaning of this field.

we are not changing the meaning of the this field. we are FIXING what's wrong in the first place. 

Explain to me please why you think it's "correct" that resources accounted for as "allocable" is not actually allocable for the scheduler.

pkg/scheduler/util.go
pdgetrf: don't we have tools to enforce this? oh right, it's reverted!

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: I do not understand what's being suggested here.

pkg/scheduler/framework/session.go
pdgetrf: being discussed elsewhere. resolved to prevent dup.

pkg/scheduler/util.go
pdgetrf: this check is already removed.

pkg/scheduler/framework/session.go
pdgetrf: this can be fixed in a following up PR which, as you already suggested, is smaller and easier to review. I do not see how this should block the current PR. It will only make this one even bigger. And we are going to address this in the starvation PR anyway. And backfill will be disabled in this PR anyway. If your purpose is to further delay the starvation PR including all the necessary testing, yes I agree we should fix it here in this PR.

pdgetrf: > What happen if Pod is patched but did not dispatch because of other error?

code has the answer.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: I'd prefer to do this in a following up PR

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: may consider removing the candidates selection all together. WIP.

pkg/scheduler/actions/backfill/backfill.go
dhzhuo: `Allocate` is defined in `session` which is meant to be reused by actions. Why should we not to use `session.Allocate` here? Can you please explain a little bit more?

pkg/scheduler/api/job_info.go
pdgetrf: WIP

pkg/scheduler/framework/session.go
pdgetrf: WIP

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: fixed

pkg/scheduler/util.go
pdgetrf: this has been removed.

pkg/scheduler/util.go
pdgetrf: will wait the tool is brought back to fix this. this shouldn't be fixed by hand.

pkg/scheduler/framework/session_plugins.go
dhzhuo: Backfill function can be enabled/disabled from configurations. `BackfillEligible` is a function provided by each plugin to decide which job is eligible to be a backfill job. What is the use case to conditionally turn on/off `BackfillEligible` function?

pkg/scheduler/api/node_info.go
dhzhuo: @k82cn Imagine kube-batch gets a starving task, and it plans to reserve all remaining idle resources on node X for the starving task. But before reserving the resources on node X, kube-batch has to make sure the allocatable resources (i.e. the total resources) of node X is bigger enough to accommodate the task. 

According to the current implementation, if node X has 5G memory and 4G memory are taken by jobs that not managed by kube-batch scheduler, node X's allocatable resource is still 5G. In this case, kube-batch may think node X has enough allocatable resource to accommodate the starving task. But the reality is different. 4G memory is occupied by tasks on which kube-batch has no control. Kube-batch cannot evict those tasks to make room for the starving task.

pkg/scheduler/framework/session_plugins.go
k82cn: > What is the use case to conditionally turn on/off BackfillEligible function?

Different plugin may provide this callback, we need a configuration to controll it.

pkg/scheduler/api/node_info.go
k82cn: > Explain to me please why you think it's "correct" that resources accounted for as "allocable" is not actually allocable for the scheduler.

It means allocatable resources on the node, not the scheduler. If we want to trace the "allocatable" resource in scheduler, please introduce a new field for it.

> In this case, kube-batch may think node X has enough allocatable resource to accommodate the starving task. 

Not, we should consider 1G memory for kube-batch; that's why `Idle`, `Used` and `Releasing` there.

dhzhuo: For starving tasks, we do not just compare a task's requesting resource with idle resource, we also compare a task's requesting resource against the total resource on the node.

pkg/scheduler/actions/backfill/backfill.go
pdgetrf: this will be re-structured

pkg/scheduler/api/node_info.go
pdgetrf: "It means allocatable resources on the node"

if it's not allocatable for the scheduler, how is it even allocatable on the node? allocatable says there's 5G memory but in fact 2G is already allocated. Why should the node still think 5G is "allocatable"?

PR #697, opened 2019-04-02
Bechmarking with kube-mark

Rajadeepan:
**What this PR does / why we need it**:
   Kubemark is integrated with kube-batch
   Test cases added to run density test on kubemark cluster
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #
https://github.com/kubernetes-sigs/kube-batch/issues/427
**Special notes for your reviewer**:

PR #810, opened 2019-04-20
Add kubemark support

TommyLike:
Add kubemark cluster script and density test

**Release note**:

```release-note

```

For #697


TommyLike:
Only conflicts are resolved and integrated into makefile, need testing on GCE to see if it's actually work.

PR #811, opened 2019-04-21
Automated cherry pick of #809: Prevent memory leak

k82cn:
Cherry pick of #809 on release-0.4.

#809: Prevent memory leak

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -482,6 +482,8 @@ func (sc *SchedulerCache) processCleanupJob() {
 		return
 	}
 
+	defer sc.deletedJobs.Done(obj)
+
 	job, found := obj.(*kbapi.JobInfo)
 	if !found {
 		glog.Errorf("Failed to convert <%v> to *JobInfo", obj)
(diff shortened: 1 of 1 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -482,6 +482,8 @@ func (sc *SchedulerCache) processCleanupJob() {
 		return
 	}
 
+	defer sc.deletedJobs.Done(obj)
```
hex108, line 485: If there is something error in line 496, we'll try to delete it in line 501. If we defer delete it, it will be deleted at last, line 501 will take no effort.

k82cn: `deleteJob` will add it to `deletedJob` queue, as `Done` is not executed, it's only marked as dirty in queue :)

k82cn: oh, there's still an issue that after `delete(sc.Jobs, job.UID)`, we should forget job in reateLimitedQueue.

@hzxuzhonghu

hzxuzhonghu: The forget  method for `DefaultControllerRateLimiter` is empty, have no effect. 

```
func (r *BucketRateLimiter) Forget(item interface{}) {
}

```

k82cn: It's better to include to align with its description in interface.

PR #812, opened 2019-04-21
Added Queue status.

k82cn:

**Release note**:
```release-note
Added number of PodGroup in different state.
```


config/crds/scheduling_v1alpha1_podgroup.yaml
```diff
@@ -41,3 +41,11 @@ spec:
           type: object
       type: object
   version: v1alpha1
+  subresources:
+    status: {}
+status:
```
hex108: We could omit it if all fields are empty?

k82cn: if omit, `UpdateStatus` will not work :)

hex108: Then it is OK. :) 

But in my CRD use case,
```
  subresources:
    status: {}
```
is enough.

k82cn: great to know, I'll remove the other part :)

k82cn: done

hex108:
LGTM

Issue #774, opened 2019-04-15
vkctrl enhancement


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**Description**:

There're several enhancements need to do for `vkctl`:

1. Add enough help message when `vkctl job -h`
1. Add `view` sub-command for job
1. Add `queue` sub-command

asifdxtreme:
cc @Rajadeepan

PR #820, opened 2019-04-22
Adding subcommand queue create and list to vkctl

Rajadeepan:
       Adding subcommand queue create and list to vkctl
Fixes # https://github.com/kubernetes-sigs/kube-batch/issues/774
image

### DIFF
--- a/pkg/cli/queue/common.go
+++ b/pkg/cli/queue/common.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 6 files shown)


pkg/cli/queue/common.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+type commonFlags struct {
+	Master        string
+	Kubeconfig    string
+	SchedulerName string
+}
+
+func initFlags(cmd *cobra.Command, cf *commonFlags) {
+	cmd.Flags().StringVarP(&cf.SchedulerName, "scheduler", "", "vn-scheduler", "the scheduler for this job")
```
hex108: `vn-scheduler` or `kube-batch`?

k82cn: should be `kube-batch`

Rajadeepan: updated with kube-batch

PR #849, opened 2019-05-13
Add Comment in allocate

thandayuthapani:
In allocate, other tasks should be considered for scheduling, even if previous task fails predicates.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -142,6 +142,8 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 
 			predicateNodes := util.PredicateNodes(task, allNodes, predicateFn)
 			if len(predicateNodes) == 0 {
+				// Further Tasks should be checked because tasks are ordered in priority, so it affects taskPriority within Job,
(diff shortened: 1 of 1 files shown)


hex108:
/lgtm

But It might affect schedule performance.

k82cn:
That'll break "task priority within job" feature :)

hex108:
> That'll break "task priority within job" feature :)

Yes, I forgot it again!!! I once had a similar PR: https://github.com/kubernetes-sigs/kube-batch/pull/646

k82cn:
> That'll break "task priority within job" feature :)
>> Yes, I forgot it again!!! I once had a similar PR: #646

:) , @thandayuthapani , would you help to add a comment on that? so the coming contributor will not open similar PR for that.

thandayuthapani:
@k82cn @hex108 Can you please review this test case https://github.com/kubernetes-sigs/kube-batch/blob/6adee18ef770f0599912340648ff01001a2f64fc/test/e2e/job.go#L331 
Because of this test case failure only this Fix was raised.  Here master priority task blocks worker priority task for allocation, but min number of tasks to be run is 1, so if worker priority is allocated, that test case passes, else tasks will not be ready at all, but that test case expects tasks to become ready.  If this fix is not necessary, then that test case has to be changed accordingly.

PR #855, opened 2019-05-17
support priorities map-reduce

wackxu:
@k82cn @hex108 

**Release note**:

```release-note
support priorities map-reduce
```

### DIFF
--- a/pkg/scheduler/util/scheduler_helper.go
+++ b/pkg/scheduler/util/scheduler_helper.go
@@ -23,10 +23,41 @@ import (
 	"sync"
 
 	"github.com/golang/glog"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/kubernetes/pkg/scheduler/algorithm"
+	schedulerapi "k8s.io/kubernetes/pkg/scheduler/api"
+	"k8s.io/kubernetes/pkg/scheduler/cache"
 
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api"
 )
+
+// HostPriority represents the priority of scheduling to a particular host, higher priority is better.
+type HostPriority struct {
+	// Name of the host
+	Host string
+	// Score associated with the host
+	Score float64
+}
+
+// HostPriorityList declares a []HostPriority type.
+type HostPriorityList []HostPriority
+
+func (h HostPriorityList) Len() int {
+	return len(h)
+}
+
+func (h HostPriorityList) Less(i, j int) bool {
+	if h[i].Score == h[j].Score {
+		return h[i].Host < h[j].Host
+	}
+	return h[i].Score < h[j].Score
+}
+
+func (h HostPriorityList) Swap(i, j int) {
+	h[i], h[j] = h[j], h[i]
+}
 
 // PredicateNodes returns nodes that fit task
 func PredicateNodes(task *api.TaskInfo, nodes []*api.NodeInfo, fn api.PredicateFn) []*api.NodeInfo {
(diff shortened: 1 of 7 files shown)


hex108 (commented):
It looks good overall. I will review it carefully later if it is not urgent.

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -55,53 +60,113 @@ func PredicateNodes(task *api.TaskInfo, nodes []*api.NodeInfo, fn api.PredicateF
...
-	var bestNodes []*api.NodeInfo
-	maxScore := -1.0
-	for score, nodes := range nodeScores {
-		if score > maxScore {
-			maxScore = score
-			bestNodes = nodes
-		}
-	}
+func SelectBestNode(priorityList schedulerapi.HostPriorityList) string {
+	sort.Sort(sort.Reverse(priorityList))
```
hex108: No need to sort them, its time complexity is `O(NlogN)`, previous is `O(N)`.

wackxu: @hex108 We need sort the HostPriorityList and select the max score node at random

hex108: We just need find MaxScores nodes, and pick one of them randomly.

wackxu: Done

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -55,53 +60,125 @@ func PredicateNodes(task *api.TaskInfo, nodes []*api.NodeInfo, fn api.PredicateF
...
+		}
+		wg.Add(1)
+		go func(index int, config algorithm.PriorityConfig) {
+			defer wg.Done()
+			if err := config.Reduce(task.Pod, nil, nodeNameToInfo, results[index]); err != nil {
+				appendError(err)
+			}
+			if glog.V(10) {
+				for _, hostPriority := range results[index] {
+					glog.Infof("%v -> %v: %v, Score: (%d)", task.Pod.Name, hostPriority.Host, config.Name, hostPriority.Score)
```
hex108: Also needs print `task.Pod.Namespace`.

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -55,53 +60,125 @@ func PredicateNodes(task *api.TaskInfo, nodes []*api.NodeInfo, fn api.PredicateF
...
 	}
-	workqueue.ParallelizeUntil(context.TODO(), 16, len(nodes), scoreNode)
-	return nodeScores
+
+	return result, nil
 }
 
 // SortNodes returns nodes by order of score
-func SortNodes(nodeScores map[float64][]*api.NodeInfo) []*api.NodeInfo {
+func SortNodes(priorityList schedulerapi.HostPriorityList, nodesInfo map[string]*api.NodeInfo) []*api.NodeInfo {
```
hex108: Score's type become `int` now. It will revert https://github.com/kubernetes-sigs/kube-batch/pull/719, does it matter? @k82cn @lmzqwer2

thandayuthapani:
Agree with @hex108 's comment https://github.com/kubernetes-sigs/kube-batch/pull/855#discussion_r285035068 , this has to be checked since this will revert changes made in https://github.com/kubernetes-sigs/kube-batch/pull/719 . Except that, Looks Good to me!!

pkg/scheduler/util/scheduler_helper.go
wackxu: Done

pkg/scheduler/util/scheduler_helper.go
wackxu: @hex108 @thandayuthapani The score calculated by priorities in k8s are all return int type now and  we do not need change it to float now.  In the future, if  priorities in k8s change the return value type, we can update correspondingly

hex108: The details is in https://github.com/kubernetes-sigs/kube-batch/issues/708.

wackxu: yes, I have saw that, but for now we have no function that return float type value. and for new function, I think it may have little affect on the result because we have many  priorities and every has its weight.  And we use some struct such as `HostPriority` from k8s. If we want to use float type, I think we should first change the type in k8s. So I think int type is enough for now. @k82cn @thandayuthapani @hex108

k82cn: We need to keep float here, and float also support int for k8s's default priorities.

wackxu: @k82cn @hex108 @thandayuthapani Updated, PTAL

hex108:
Thanks! LGTM.

@wackxu could you please squash the commits?

wackxu:
@k82cn @hex108 Rebased, PTAL

PR #859, opened 2019-05-25
update node info no matter what node info changed

wackxu:
we  should update node info no matter what node info changed. for now, if node Conditions changed, we do not update the cache and pod may schedule on node that was not ready. 

/assign @k82cn @hex108

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -18,7 +18,6 @@ package cache
 
 import (
 	"fmt"
-	"reflect"
 
 	"github.com/golang/glog"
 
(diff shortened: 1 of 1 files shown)


pkg/scheduler/cache/event_handlers.go
```diff
@@ -269,19 +268,10 @@ func (sc *SchedulerCache) addNode(node *v1.Node) error {
 	return nil
 }
 
-func isNodeInfoUpdated(oldNode, newNode *v1.Node) bool {
-	return !reflect.DeepEqual(oldNode.Status.Allocatable, newNode.Status.Allocatable) ||
-		!reflect.DeepEqual(oldNode.Spec.Taints, newNode.Spec.Taints) ||
-		!reflect.DeepEqual(oldNode.Labels, newNode.Labels) ||
-		!reflect.DeepEqual(oldNode.Spec.Unschedulable, newNode.Spec.Unschedulable)
```
hex108, line 276: Is it OK to add `!reflect.DeepEqual(oldNode.Status.Conditions, newNode.Status.Conditions)` here? There are a lot of NodeUpdate, and we do not need update it every time.

k82cn: Check condition is better :)

wackxu: In default scheduler, no matter what node info changed, it always update the cache.  There are  not only node condition that we cared about when scheduling but also some other info, such as when we add `ImageLocalityPriority`, we need care about the change of node status Images info. I do not think it will have any benefits if we add more and more `DeepEqual` than we update the cache directly when node update.

k82cn: hm... let's get this merged firstly; if any performance issue, let's enhance it in another PR.

k82cn: @hex108 , WDYT?

hex108: OK with it.

PR #860, opened 2019-05-25
Fix wrong calculation for nodeinfo used

wackxu:
when node info is updated and set node, we should reset the `Used` to EmptyResource otherwise the Used is calculated twice when update node

/assign @k82cn @hex108

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -90,6 +90,7 @@ func (ni *NodeInfo) SetNode(node *v1.Node) {
 	ni.Allocatable = NewResource(node.Status.Allocatable)
 	ni.Capability = NewResource(node.Status.Capacity)
 	ni.Idle = NewResource(node.Status.Allocatable)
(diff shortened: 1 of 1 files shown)


pkg/scheduler/api/node_info.go
```diff
@@ -90,6 +90,7 @@ func (ni *NodeInfo) SetNode(node *v1.Node) {
 	ni.Allocatable = NewResource(node.Status.Allocatable)
 	ni.Capability = NewResource(node.Status.Capacity)
 	ni.Idle = NewResource(node.Status.Allocatable)
+	ni.Used = EmptyResource()
```
k82cn, line 93: If `ni.Node == nil`, we will not update `ni.Used` in `AddTask`; so it seems not necessary to reset `ni.Used`.

k82cn: we need to handle node update case.

wackxu: Yes, that is what I mean

Issue #866, opened 2019-06-28
Support both version of PodGroup


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

We also need to support `PodGroup.scheduling.sigs.dev/v1alpha2` in https://github.com/kubernetes-sigs/kube-batch/tree/master/pkg/apis/scheduling/v1alpha2 .

k82cn:
/assign @asifdxtreme 

BTW, we also need to include Queue in v1alpha2.

thandayuthapani:
@k82cn Do we need to have two different packages and build two different binaries for different versions.  Or we have to support both the version in a single package and have it in single binary(In this approach, we need to have two objects of podGroup of different versions to be maintained in JobInfo struct)?

k82cn:
just one binary; introduce a new internal object PodGroupInfo, convert different version of podGroup into internal object.

thandayuthapani:
@k82cn Even when we convert those types to internal object, we are facing problems when event is to be generated to that object, since it will not be runtime object.  
Even if we implement deep copy function to that internal object to be accepted as runtime object(needs deep copy function implemented on those structs), then API server will not be able to find that type since that is internal object.

PR #870, opened 2019-07-02
Support Both versions of PodGroup

thandayuthapani:
Add v1alpha2 PodGroup version in kube-batch, so user can create PodGroup in v1alpha2 version and kube-batch can schedule pods related to that PodGroup.
**Which issue(s) this PR fixes**:
https://github.com/kubernetes-sigs/kube-batch/issues/866


k82cn:
would you help to also add Queue into `scheduling.sig.dev` ?

thandayuthapani:
> would you help to also add Queue into `scheduling.sig.dev` ?

I can raise different PR for that, if that is okay.

PR #889, opened 2019-08-30
fix reclaim ping-pong

suigh:
There is a ping-pong for reclaim.

The env:
1. build a k8s cluster with 2 hosts, each host has 8 cpu.
2. build a job (job1) with 1 container for 6 cpu, and another job (job2) in another queue with 1 container for 6 cpu.
3. job2 will reclaim job1.
4. the reclaimed res is assigned to job1 again.

step 3 and 4 will be repeated, ping-pong happens.

The current reclaim behavior is wrong, and there is similar error for preemption. Let's fix them step by step, just handle the reclaim case here.

In this case, the changed parts are:
1: Ignore the master host when calculating the deserved resource since it cannot be used at all. With this, we can get the correct deserved resource.
2: Change the plug-in behavior for reclaim. Reclaim will not happen unless it passes all the plug-in's check. This make sense to me as if reclaim can happen when it can pass part of the plug-ins, this will break the rule of other plug-ins (it is proportion in this case), and the broken plug-in will break the reclaim behavior later, this is why ping-pong happens.

**Release note**:
```release-note
Reclaim ping-pong issue resolved.
```

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -22,7 +22,7 @@ import (
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api"
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/api/helpers"
 	"github.com/kubernetes-sigs/kube-batch/pkg/scheduler/framework"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/kubernetes/pkg/scheduler/cache"
 )
 
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -66,7 +66,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		ignore := false
 		for i := range taints {
 			t := taints[i]
-			if(t.Effect == v1.TaintEffectNoSchedule || t.Effect == v1.TaintEffectNoExecute) {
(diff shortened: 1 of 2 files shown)


suigh:
@k82cn There is following error from "Travis CI build", seems a system error, how can we fix it?

"dpkg-deb: error: archive '/var/cache/apt/archives/kubectl_1.15.3-00_amd64.deb' has premature member 'control.tar.xz' before 'control.tar.gz', giving up
dpkg: error processing archive /var/cache/apt/archives/kubectl_1.15.3-00_amd64.deb (--unpack):
 subprocess dpkg-deb --control returned error exit status 2
Errors were encountered while processing:
 /var/cache/apt/archives/kubectl_1.15.3-00_amd64.deb"

Thanks

k82cn:
@suigh , I restarted it :)

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -56,8 +58,24 @@ func (pp *proportionPlugin) Name() string {
...
-	// Prepare scheduling data for this session.
+	// Prepare scheduling data for this session, the nodes that cannot be scheduled will be ignored
 	for _, n := range ssn.Nodes {
+		nodeInfo := cache.NewNodeInfo(n.Pods()...)
+		nodeInfo.SetNode(n.Node)
+		taints, _ := nodeInfo.Taints()
+		ignore := false
+		for i := range taints {
+			t := taints[i]
+			if t.Effect == v1.TaintEffectNoSchedule || t.Effect == v1.TaintEffectNoExecute {
```
k82cn, line 69: this is handled in predicates plugins :)

PR #895, opened 2019-10-23
added priority based preemption to priority plugin

### LINKED ISSUE #894: Preempt tasks with lower priority

/kind feature

**Description**:

You should be able to configure `kube-batch`, so it preempts tasks of lower priority jobs when higher priority job arrives. Based on the code I read and what I observed on the cluster it looks like there is no way to enforce that. It looks like only `gang`, `drf` and `conformance` provide "preemptable functions". 

Possible solution is to add preemptable function to `priority` plugin:

```go
preemptableFn := func(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) []*api.TaskInfo {
	preemptorJob := ssn.Jobs[preemptor.Job]
	var victims []*api.TaskInfo
	for _, preemptee := range preemptees {
		preempteeJob := ssn.Jobs[preemptee.Job]
		if preempteeJob.Priority >= preemptorJob.Priority {
			glog.V(3).Infof("Can not preempt task <%v/%v> because "+
				"preemptee has greater or equal job priority (%d) than preemptor (%d)",
				preemptee.Namespace, preemptee.Name, preempteeJob.Priority, preemptorJob.Priority)
		} else {
			victims = append(victims, preemptee)
		}
	}
	glog.V(3).Infof("Victims from Priority plugins are %+v", victims)
	return victims
}
ssn.AddPreemptableFn(pp.Name(), preemptableFn)
```

Let me know if this make sense.

mateuszlitwin:
Priority plugin now adds preemptable function which makes preemptee a victim if and only if job priority of the preemptee is strictly lower than job priority of the preemptor.

This is consistent with default behavior of the standard kubernetes scheduler: if a Pod cannot be scheduled, the scheduler tries to preempt (evict) lower priority Pods to make scheduling of the pending Pod possible (see https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/).

To disable this functionality use `enablePreemptable: false` in the `priority` plugin configuration.

Closes kubernetes-sigs/kube-batch#894

**Which issue(s) this PR fixes** :
Fixes #894

**Release note**:

```release-note
Priority plugin now adds preemptable function which makes preemptee a victim if and only if job priority of the preemptee is strictly lower than job priority of the preemptor. To disable this functionality use `enablePreemptable: false` in the `priority` plugin configuration.
```

### DIFF
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -77,14 +77,14 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
 			preemptable := job.MinAvailable <= occupid-1 || job.MinAvailable == 1
 
 			if !preemptable {
-				glog.V(3).Infof("Can not preempt task <%v/%v> because of gang-scheduling",
+				glog.V(4).Infof("Can not preempt task <%v/%v> because of gang-scheduling",
 					preemptee.Namespace, preemptee.Name)
 			} else {
 				victims = append(victims, preemptee)
 			}
 		}
 
-		glog.V(3).Infof("Victims from Gang plugins are %+v", victims)
+		glog.V(4).Infof("Victims from Gang plugins are %+v", victims)
 
 		return victims
 	}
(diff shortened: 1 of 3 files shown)


mateuszlitwin:
Adjusted log level from 3 to 4 inside all preemptable functions for consistency, to make build pass and to not spam too many logs.

Issue #891, opened 2019-09-12
pod will stuck in pending status when host down


davidstack:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
 /kind bug
> /kind feature

**What happened**:
pod stuck in pending status after bind to the host which is powered off
**What you expected to happen**:
pod will rescheduled when host is down
**How to reproduce it (as minimally and precisely as possible)**:
because of the cluster snapshot, host status will not update immediately,so pod may be binded to a host which is powered off in this schedule period. but kube-batch will update this pod in Binding status. but actually, the pod is in pending status

**Anything else we need to know?**:
i think kube-batch should one goroutine to check this status ,and change the status of pod from binding to pennding in kube-batch cache. 
**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

k82cn:
Will GC delete the pod to trigger re-scheduling?

davidstack:
the node has been poweroff，kubelet can not gc ?

PR #909, opened 2019-11-20
revert cache to original state on evict and bind errors

mateuszlitwin:
This change ensures that node, task and job info will remain unchanged in case of an error during SchedulerCache.Bind and SchedulerCache.Evict calls.
Before if error occurred during binding phase (e.g. "Selected node NotReady") task could get stuck in the Binding status indefinitely while the real pod would be in the Pending status.

- SchedulerCache.Evict and SchedulerCache.Bind will revert task status on NodeInfo.UpdateTask and NodeInfo.AddTask errors.
- Modified behavior of NodeInfo.AddTask. AddTask will now update task's node name upon successful addition, this is similar to how JobInfo.UpdateTaskStatus updates task's status.
- Handling unchecked error in JobInfo.UpdateTaskStatus.
- FATAL logging in NodeInfo.UpdateTask when impossible situation happens - failing to add a task after removal of a task from node info.

Might be related to #891.

**Release note**:

```release-note

```

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -163,15 +163,18 @@ func (ni *NodeInfo) allocateIdleResource(ti *TaskInfo) error {
 		ni.Idle.Sub(ti.Resreq)
 		return nil
 	}
-	ni.State = NodeState{
-		Phase:  NotReady,
-		Reason: "OutOfSync",
-	}
 	return fmt.Errorf("Selected node NotReady")
 }
 
 // AddTask is used to add a task in nodeInfo object
+//
+// If error occurs both task and node are guaranteed to be in the original state.
 func (ni *NodeInfo) AddTask(task *TaskInfo) error {
+	if len(task.NodeName) > 0 && len(ni.Name) > 0 && task.NodeName != ni.Name {
+		return fmt.Errorf("task <%v/%v> already on different node <%v>",
+			task.Namespace, task.Name, task.NodeName)
+	}
+
 	key := PodKey(task.Pod)
 	if _, found := ni.Tasks[key]; found {
 		return fmt.Errorf("task <%v/%v> already on node <%v>",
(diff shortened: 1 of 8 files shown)


k82cn:
Good catch ! Please ping me when it's ready for review :)

mateuszlitwin:
@k82cn Will do.

BTW. The bug that happened on my cluster was that each time "Selected node NotReady" happened during allocation binding, the task would end up in the Binding status even though pod was Pending. This PR is attempt to make Bind/Evict handle failures like that.

However I do not understand why "Selected node NotReady" would happen in the first place. It seems to me it can only happen if `task.InitResreq.LessEqual(node.Idle)` and `node.Idle.Less(task.Resreq)`. However I would expect that `task.Resreq.LessEqual(task.InitResreq)` is always true? (https://github.com/kubernetes-sigs/kube-batch/blob/master/pkg/scheduler/actions/allocate/allocate.go#L160; https://github.com/kubernetes-sigs/kube-batch/blob/master/pkg/scheduler/api/node_info.go#L162)

EDIT: I suppose it was some race condition (SchedulerCache and Session having different nodes)

pkg/scheduler/api/node_info.go
```diff
@@ -163,15 +163,18 @@ func (ni *NodeInfo) allocateIdleResource(ti *TaskInfo) error {
 		ni.Idle.Sub(ti.Resreq)
 		return nil
 	}
-	ni.State = NodeState{
```
mateuszlitwin, line 166: Removing this because I do not want to modify node info state on error, also this state might not be true depending on the context.

I think, if it is necessary, this state could be set explicitly after receiving an error from `AddTask()`, depending on the context in which error occurred.

mateuszlitwin:
@k82cn Ready for review.

k82cn:
> Ready for review.

Thanks very much! I'm going to check that :)

Issue #907, opened 2019-11-17
Statement unevict method doesn't update nodeInfo.Releasing field as expected


yodarshafrir1:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> kind bug

**What happened**:
`unevict` method in file: `statement.go` doesn't reset the field `Releasing` of the task's nodeInfo to the expected value (the value of `Releasing` which was changed by `evict` method will remain as if the task was released from the node).

This means that when a task is running on a node, `evict` will be called and then `unevict` will be called, the Releasing value of the current nodeInfo will be as if the task was evicted from the node.

This happens because `unevict` calls `AddTask`. In file: `node_info.go` `AddTask` there's a condition to check if the task exists on this node and if so we return an error and don't set the `Releasing` value to the right value.

Instead the method `unevict` in file: `statement.go` should call `UpdateTask` (this is the fix). In `UpdateTask` the `Releasing` value will be updated as if the task was never evicted from the node.

**What you expected to happen**:
The `Releasing` value of the node should be as if the task was never evicted from its node.

**How to reproduce it (as minimally and precisely as possible)**:
Create a use case where `preempt` evicts a task but fails to pipeline a task, then `statement.Discard` will be called.
Print the value `Releasing` field of the nodeinfo and see that it has a wrong value.

k82cn:
is there anyway to trigger this case by workload?

yodarshafrir1:
This happened to me after changing the order of the actions, so `preempt` will run before `allocate`.

When `preempt` will run, we just need an attempt to `evict` a task and yet, not have enough resources to `pipeline` the preemptor. On this case `stmt.Discard()` is called.
When `Allocate` will run after this case, due to the following `if`:

```
// Allocate releasing resource to the task if any.
task.InitResreq.LessEqual(node.Releasing) {
```

A pending task that fits the `node.Releasing` will be pipelined to the node, even though that the running task was never really evcited.


By the way, reading the comments in the code suggests that it was intended to update the task and not add a task.

```
// Update task in node.
if node, found := s.ssn.Nodes[reclaimee.NodeName]; found {
    node.AddTask(reclaimee)
}
```

k82cn:
thanks very much for your contribution :)

yodarshafrir1:
Issue was solved in: https://github.com/kubernetes-sigs/kube-batch/pull/912

PR #912, opened 2019-11-26
Change Statement unevict method to call UpdateTask

yodarshafrir1:
**What this PR does / why we need it:**
This change ensures that when `unevict` is called the value of `nodeInfo.Releasing` will be updated as if the task was evicted.

Before this fix, because the task exists on the current `nodeInfo`, every time `unevict` is called the `Releasing` value of the nodeInfo isn't updated and the value remains same as if the task was `evicted`

**Why does it happen:**
This happens because `unevict` calls `AddTask`:
```
// Update task in node.
if node, found := s.ssn.Nodes[reclaimee.NodeName]; found {
     node.AddTask(reclaimee)
} 
```
`AddTask` will throw an error because this task already exists under the current nodeInfo and therefore will not update the `Releasing` field as if the current task is `Running`

Issue:  [#907](https://github.com/kubernetes-sigs/kube-batch/issues/907)

### DIFF
--- a/pkg/scheduler/framework/statement.go
+++ b/pkg/scheduler/framework/statement.go
@@ -95,7 +95,7 @@ func (s *Statement) unevict(reclaimee *api.TaskInfo, reason string) error {
 
 	// Update task in node.
 	if node, found := s.ssn.Nodes[reclaimee.NodeName]; found {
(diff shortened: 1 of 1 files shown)


yodarshafrir1:
I signed it

PR #924, opened 2020-01-05
Document framework, plugins and actions

### LINKED ISSUE #887: [Question] Confusing documentation: actions, plugins and where to start?
First of all, I'm aware that I should do this kind of requests in StackOverflow or in Slack, but I haven't found any kube-batch questions in any of both.

To explain a bit my context, we are using Argo as a system to schedule and launch (batch) pods in our K8s cluster. However, the default k8s scheduler uses a round-robin system to deploy the pods through nodes. Also, we use a node autoscaler to scale up the nodes in our system. Eventually, what happens is that we end with lots of nodes with one container each, and we would expect that nodes would be filled before scheduling in a new node, so empty nodes can get evicted safely from the cluster. Node autoscaler cannot evict nodes with jobs running, because they may stop the batch process in the middle...

I guess that one of the aims in this kube-batch is to support this kind of scheduling, but after few hours around the documentation, I'm quite confused. In the tutorial, there are mentions on actions and plugins. What are the differences between them, and how do they interact? Also, what does each plugin or action?

I've ended reading Go code to understand what is the backfilling action logic and the conformance plugin, but I'm not even sure what each of them does or how are they used. Maybe I'm new in all the scheduling jargon, but I do not really understand (or been able to find in the documentation) how this project works. Neither I've found help in StackOverflow and in Slack.

I'd like to help in documentation if needed, but also I'm not sure if this project covers my needs or not, so if somehow can help me, I'd be willing to give a hand in documentation, because at least for me, I do not know where to start aside of the 2 small samples in the tutorial (and I do not need gang-scheduling nor pod priority scheduling at the moment....)

Thanks a lot in advance, and I hope I can give a hand in the project!

ingvagabund:
Giving more detailed description of what individual pieces are and how their interact with each other. So it's easier to orient in the code. The document is not meant to be a full description of the project.

Fixes: #887

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/doc/design/framework.md
+++ b/doc/design/framework.md
@@ -1,0 +1,246 @@
+# Kube-batch scheduler framework
+
+@ingvagabund; Jan 7, 2020
+
+Kube-batch scheduler provides a framework which allows to simultaneously
+schedule a group of tasks (e.g. all tasks must be first scheduled to available nodes
+before each task starts to run). By integrating multi level queue scheduling
+mechanism it allows to categorize and prioritize jobs (jobs in higher priority
+queues will get processed first before jobs in lower priority queues).
+Extensible through plugins the framework allows to define various conditions
+for when a task can be preempted, what it means for a job to be ready, how to
+sort jobs/tasks, how to prioritize multi-level queues, etc.
+Logic responsible for the scheduling itself (including preemption) is implemented
+through actions which build on top of the framework and use logic provided by the plugins.
+
+## Framework
+
+The framework provides various methods to be used within actions. E.g. `Reclaimable`,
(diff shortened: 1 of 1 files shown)


k82cn:
thanks very much on doc improvement :)

k82cn (approved):

ingvagabund:
> thanks very much on doc improvement :)

@k82cn  Happy to help. I think it's ready for review. There were some places where I was not sure so the explanation may sound inaccurate. I am open to suggestion of improvement.

ingvagabund:
Depends on https://github.com/kubernetes-sigs/kube-batch/pull/926

PR #944, opened 2020-08-06
Avoid pushing empty jobs into queue

lixiang233:
Currently we may push am empty job back to its queue and that may slightly affect the performance.


riverzhang:
LGTM

hzxuzhonghu:
@lixiang233 Can you help fix that in https://github.com/volcano-sh/volcano

